PR #18, opened 2023-02-20
Update heat-operator

bshephar:
This patch fixes various implementation issues that have since been outdated or deprecated. It also implements some fixes that allow the init container job to launch and configure the DB service.

The second commit adds required changes for Heat API and Heat Engine


templates/heatapi/config/heatapi-config.json
```diff
@@ -0,0 +1,43 @@
...
+      "preserve_properties": true,
+      "perm": "0644"
+    },
+    {
+      "source": "/usr/bin/heat-wsgi-api",
+      "dest": "/var/www/cgi-bin/heat/heat_api",
+      "merge": false,
+      "preserve_properties": true,
+      "perm": "0644"
+    }
```
kajinamit, line 24: probably we can use https://review.opendev.org/c/openstack/tripleo-common/+/874526 ?

bshephar: When we build the container to begin with you mean?

kajinamit: Yes that is correct.

Giving it another though I guess we can use /usr/bin/ script instead of copying it to /var/lib/cgi-bin. The current symlink comes from what was done by puppet.

See https://github.com/openstack-k8s-operators/nova-operator/blob/master/templates/novaapi/config/httpd.conf#L42

bshephar: Since selinux is disabled in the container, we can probably fairly easily use the script from /usr/bin. I'll change the implementation and push a new commit to this PR before we merge.

PR #21, opened 2023-03-16
Use time.Duration for timeout

kajinamit:
instead of int. This allows us to use more fine grained timeout value such as 0.1, and encodes the unit in the type itself.

Note the current version of lib-common and keystone-operator in go.mod contains the change in these dependent libraries.


bshephar (approved):
Thanks, sounds reasonable.

PR #30, opened 2023-03-16
Fix wrong project name in workflows

kajinamit:
It seems I used the wrong sed command and some files were not updated properly after imported from horizon-operator.


kajinamit:
Still testing ...

kajinamit:
I still have to look into the failure in gotests but at least lint jobs are all passing

kajinamit:
I'll work on fixing the gotests separately. Let's merge this to avoid messing up horizon thing.

PR #41, opened 2023-03-23
Add user for stack_domain_admin

bshephar:
This change adds the stack_domain_admin user and password to keystone, while also updating the heat.conf template with the relevant data.


controllers/heat_controller.go
```diff
@@ -373,6 +374,42 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
...
+	keystoneAPI, err := keystonev1.GetKeystoneAPI(ctx, helper, instance.Namespace, map[string]string{})
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	// get the password of the service user from the secret
+	password, _, err := oko_secret.GetDataFromSecret(
+		ctx,
+		helper,
+		instance.Spec.Secret,
+		10,
```
kajinamit, line 386: should be time.Duration(10)

controllers/heat_controller.go
```diff
@@ -373,6 +374,42 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
...
+	os, _, err := keystonev1.GetAdminServiceClient(
+		ctx,
+		helper,
+		keystoneAPI,
+	)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	_, err = os.CreateUser(
```
kajinamit, line 403: we also have to create heat_stack domain and assign admin role between heat_stack and this heat_stack_domaiN_admin user

controllers/heatengine_controller.go
```diff
@@ -455,9 +455,23 @@ func (r *HeatEngineReconciler) generateServiceConfigMaps(
...
+		return err
+	}
+
 	templateParameters := map[string]interface{}{
-		"KeystonePublicURL": authURL,
-		"ServiceUser":       instance.Spec.ServiceUser,
+		"KeystonePublicURL":   authURL,
+		"ServiceUser":         instance.Spec.ServiceUser,
+		"StackDomainAdmin":    heat.StackDomainAdmin,
+		"StackDomainPassword": password,
```
kajinamit, line 474: StackDomainAdminPassword. Also this should be part of secret.

controllers/heatengine_controller.go
```diff
@@ -455,9 +455,23 @@ func (r *HeatEngineReconciler) generateServiceConfigMaps(
...
+	if err != nil {
+		return err
+	}
+
 	templateParameters := map[string]interface{}{
-		"KeystonePublicURL": authURL,
-		"ServiceUser":       instance.Spec.ServiceUser,
+		"KeystonePublicURL":   authURL,
+		"ServiceUser":         instance.Spec.ServiceUser,
+		"StackDomainAdmin":    heat.StackDomainAdmin,
```
kajinamit, line 473: StackDominAdminUsername would be better

controllers/heat_controller.go
bshephar: I think I'll need to add a new Domain package to lib-common to do this. Rather than manually using doing it here. I'll work on the lib-common patch for it.

controllers/heatengine_controller.go
bshephar: Oh, actually. This isn't quite resolved. You mean we need to add these to the osp-secret secret?

kajinamit: Because this is password, we should not put it into ConfigMap. Probably adding this to osp-secret is the right way to go, because we generally keep passwords for all keystone users there.

controllers/heat_controller.go
bshephar: For reference:
https://github.com/openstack-k8s-operators/lib-common/pull/224

stuggi: where do we plan to delete this user? I think for now its ok. do do it like this. I was planning to split the keystoneservice CRD as this right now creates the user and the service. we probably want to have a crd for user and service so that we use the same interface. maybe we then should have there also a keystonedomain crd? And like we do with the service and endpoint the user/domain then gets cleaned up. Does that make sense?

kajinamit (approved):
LGTM. We might want a separate password for heat_domain_admin but we can revisit it later.

PR #45, opened 2023-03-27
Wrap errors in return

bshephar:
Golang errors should be wrapped in return statements to make life easier while troubleshooting where errors are coming from. This change adds fmt.Errorf() wrapping to each error along with a more meaningful error message for each.

### DIFF
--- a/kuttl-test.yaml
+++ b/kuttl-test.yaml
@@ -19,7 +19,7 @@ kind: TestSuite
 reportFormat: JSON
 reportName: kuttl-test-heat
 namespace: openstack
-timeout: 450
+timeout: 400
 parallel: 1
 suppress:
   - events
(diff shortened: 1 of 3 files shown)


controllers/heat_controller.go
```diff
@@ -350,23 +350,23 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 	// Handle service init
 	ctrlResult, err := r.reconcileInit(ctx, instance, helper, serviceLabels)
 	if err != nil {
-		return ctrlResult, err
+		return ctrlResult, fmt.Errorf("Error initializing service: %w", err)
```
kajinamit, line 353: I think we should use the consistent message in a similar logic.

In HeatAPI/Engine the error message for reconcieInit failure is `fmt.Errorf("Failed to initialize %s: %w", instance.Name, err)` which is different from this one.

Also

PR #53, opened 2023-04-03
Update operator metadata

kajinamit:


bshephar:
Maybe `make manifests` again to regain the CRD yaml file?

kajinamit:
Seems I unintentionally removed the `links` key, which is fixed in the latest version.

PR #55, opened 2023-04-06
Extend timeout in golangci-lint

kajinamit:
The command sometimes fails because of timeout (default 1m0s) in CI. Let's increase timeout to avoid frequent failure.


kajinamit:
I'll ask feedback from CI works to discuss whether we can distribute this to all operators.

bshephar (approved):

PR #65, opened 2023-04-13
Squash config files into a single layer

kajinamit:
... instead of creating layer per service. This allows us to maintain all options in a single heat.conf file instead we maintain the config file for each service. So far we are not aware of any specific requirement to tune options differently in each service by default. Users can use custom config interface in Api/Engine CR in case they need to tune a parameter for a specific service.


controllers/heat_controller.go
```diff
@@ -713,12 +715,7 @@ func (r *HeatReconciler) generateServiceConfigMaps(
 			Labels:        cmLabels,
 		},
 	}
-	err = configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
-	if err != nil {
-		return nil
-	}
-
-	return nil
+	return configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
```
bshephar, line 718: Ah nice catch on the unnecessary error handling there too.

bshephar (approved):
Nice, this looks good to me. Glad to have a more simple implementation.

controllers/heat_controller.go
kajinamit: Actually the current implementation is wrong because it completely ignores the failure.

PR #68, opened 2023-04-17
Create a secret to store auth_encryption_key

kajinamit:
This introduces a k8s secret to store the [DEFAULT] auth_encryption_key option. The option is rendered in a separate config file in the conf.d directory.


kajinamit:
We have to discuss a few more points before we finalize this implementation.

- Ideally we should move all the other secret options (such as password) to the separate file. Do we agree with the current implementation(to all)
- If we agree we create a separate file for all secret options, then how we design the secrets creates ? Current implementation does not really allow users to give auth_encryption_key. Do we need a separate secret or should we add it to the global secret ?

PR #79, opened 2023-04-21
Disable log file and use stderr for log output

kajinamit:
This change disable usage of log file and make all log outputs sent to stderr. The file is created inside pod and can fill up container file system.


bshephar (approved):
/lgtm 

nice catch.

PR #82, opened 2023-04-24
Wrap errors when returning them

bshephar:
Golang errors should be wrapped in return statements to make life easier while troubleshooting where errors are coming from. This change adds fmt.Errorf() wrapping to each error along with a more meaningful error message for each.


abays (approved):
/lgtm

PR #84, opened 2023-04-24
Remove unused transportURLSecret from heatAPI/Engine CR

kajinamit:
The transportURLSecret spec is no longer used by these two CRs when we merged config file generation to the base implementation in the heat CR.

This also removes a few invalid/outdated comments in init.sh script.


kajinamit:
Hmm. it seems cinder-operator implements transportURLSecret in all CRs and, for sub CRs the operator makes sure that the secret exists. Should we implement the same logic instead of removing this spec from sub CRs ?

kajinamit:
> Hmm. it seems cinder-operator implements transportURLSecret in all CRs and, for sub CRs the operator makes sure that the secret exists. Should we implement the same logic instead of removing this spec from sub CRs ?

This is not really required because heatAPI/Engine requires config map generated by heat, which is dependent on transport url secret.

kajinamit:
ok it turned out we need this to run init container for api/engine...

PR #94, opened 2023-05-03
Webhooks for default images

abays:
Default container images are now handled via env vars and internal webhooks, as per the pattern we are using for all NextGen operators.


api/v1beta1/heat_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
+//+kubebuilder:webhook:path=/validate-heat-openstack-org-v1beta1-heat,mutating=false,failurePolicy=fail,sideEffects=None,groups=heat.openstack.org,resources=heats,verbs=create;update,versions=v1beta1,name=vheat.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &Heat{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Heat) ValidateCreate() error {
+	heatlog.Info("validate create", "name", r.Name)
+
+	// TODO(user): fill in your validation logic upon object creation.
```
bshephar, line 86: Since we're just using the Defaulter interface here, we don't really mind about Validator for now, let's just remove these default `TODO`'s here. We can add logic to these later, but `user` already has enough work TODO in this repo as it is.

bshephar (approved):
lgtm. Be nice to remove those generic todo's though, I already have a whole bunch of them that I need to find and clean up.

PR #95, opened 2023-05-04
Standardize Makefile and kustomize image refs

bshephar:


config/manager/kustomization.yaml
```diff
@@ -1,10 +1,16 @@
 resources:
-- manager.yaml
+  - manager.yaml
```
kajinamit, line 2: should we remove the first indent to make the file consistent with the other operators ?

config/manager/kustomization.yaml
```diff
@@ -1,10 +1,16 @@
...
 
 generatorOptions:
   disableNameSuffixHash: true
 
 configMapGenerator:
-- name: manager-config
-  files:
-  - controller_manager_config.yaml
+  - files:
```
kajinamit, line 8: ditto

config/manager/kustomization.yaml
```diff
@@ -1,10 +1,16 @@
...
+      - controller_manager_config.yaml
+    name: manager-config
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+  - name: controller
```
kajinamit, line 14: ditto

PR #102, opened 2023-05-11
Ensure status is updated correctly after reconcile

bshephar:
We currently missing updating the status field of the CR once the resources are reconciled. This patch adds the required update.


bshephar:
Note that there is two different approached being used here. So we need to decide on one or the other before merging this. The other option for this is to use `r.Client.Status().Update()` which we do here:
https://github.com/bshephar/heat-operator/blob/main/controllers/heat_controller.go#L570

kajinamit:
I checked the other operators/controllers and it looks like the issue is caused by the current implementation in `defer func`. I've updated it according to the other operators and now status is properly patched.

https://github.com/openstack-k8s-operators/heat-operator/pull/103

PR #106, opened 2023-05-15
Amend some fields heat kuttl tests

cescgina:
Some fields in the assert when running the kuttl tests and are causing
them to fail. This change also modifies the directory structure of the
kuttl tests, bringing the common folder one level up, otherwise kuttl
sees it as another test case.

### DIFF
--- a/tests/kuttl/common/assert-sample-deployment.yaml
+++ b/tests/kuttl/common/assert-sample-deployment.yaml
@@ -59,39 +59,9 @@ spec:
   serviceUser: "heat"
 status:
   databaseHostname: openstack
-  heatAPIReadyCount: 1
-  heatConductorReadyCount:
-    null_conductor_group_null: 1
----
-apiVersion: heat.openstack.org/v1beta1
-kind: HeatAPI
-metadata:
-  finalizers:
-    - HeatAPI
-  name: heat-api
-  namespace: openstack
-  ownerReferences:
-    - apiVersion: heat.openstack.org/v1beta1
-      blockOwnerDeletion: true
-      controller: true
-      kind: Heat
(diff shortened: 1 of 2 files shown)


tests/kuttl/common/assert-sample-deployment.yaml
```diff
@@ -123,8 +119,6 @@ spec:
   secret: osp-secret
   serviceUser: heat
   transportURLSecret: rabbitmq-transport-url-heat-heat-transport
-status:
```
kajinamit, line 126: I checked the other operators (eg. keystone-operator) and readyCount is asserted. Should we keep this ? also we have the same field for HeatEngine CR and can be asserted.

tests/kuttl/common/assert-sample-deployment.yaml
```diff
@@ -59,9 +59,6 @@ spec:
   serviceUser: "heat"
 status:
   databaseHostname: openstack
-  heatAPIReadyCount: 1
```
kajinamit, line 62: I think it makes sense to keep heatApiReadyCount . We can also assert the following two fields.

heatEngineReadyCount: 1
transportURLSecret: rabbitmq-transport-url-heat-heat-transport

cescgina: You're right, before for some reason these fields were not exposed, but now they appear, I'll put them back

tests/kuttl/common/assert-sample-deployment.yaml
cescgina: added both of them

bshephar (approved):

PR #108, opened 2023-05-16
Fix unused/missing conditions of Heat CR

kajinamit:


kajinamit:
I've tested this PR in my local and confirmed all conditions become ready.

bshephar (approved):
Ahh Conductor still there too. 

lgtm

PR #114, opened 2023-05-19
Remove spec keys copied from Heat CR

kajinamit:
Some spec keys of sub CRs(eg. HeatAPI) are overridden by the same keys of the Heat CR, so defining these keys has no effect.


kajinamit:
@bshephar we ideally should remove these keys from spec(as we discussed earlier today) but in the mean time we can remove these keys from samples at least.

bshephar (approved):

PR #116, opened 2023-05-19
Rename spec.APIEndpoint to plural

kajinamit:
... because the field contains multiple endpoints(heat-api and heat-api-cfn, internal and public)


abays (approved):
Looks like we have similar copy-paste mistakes elsewhere, such as [1].

/lgtm

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/880c11fddd0393b39ce82f53bae62114b96a0669/api/v1beta1/glanceapi_types.go#L93

PR #121, opened 2023-05-23
Make sure conditions for keystone resources are initialized

kajinamit:
This makes sure that the conditions describing status of keystone resource creation are initialized.


abays (approved):
/lgtm

PR #128, opened 2023-05-29
Remove redundant if blocks

kajinamit:
The returned values do not change even if we remove these if blocks.


bshephar (approved):
True, nice one.

PR #133, opened 2023-06-01
Add back secret keys for sub CRs

kajinamit:
These keys are required because we inherit definition of sub CRs. Currently this does not appear as a problem because webhook is modifying the value, but we should fix that behavior ideally and then the problem appears
    
This adds the keys back until we determine how we handle spec keys inherited from top-level CR.

### DIFF
--- a/config/samples/heat_v1beta1_heat.yaml
+++ b/config/samples/heat_v1beta1_heat.yaml
@@ -18,6 +18,7 @@ spec:
       service: false
     replicas: 1
     resources: {}
+    secret: osp-secret
   heatCfnAPI:
     containerImage: "quay.io/podified-antelope-centos9/openstack-heat-api-cfn:current-podified"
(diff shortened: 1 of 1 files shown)


bshephar (approved):

bshephar (commented):
What about passwordSelectors? Do we experience the same issue there?
https://github.com/openstack-k8s-operators/heat-operator/blob/main/pkg/heatapi/deployment.go#L152-L155

kajinamit:
> What about passwordSelectors? Do we experience the same issue there? https://github.com/openstack-k8s-operators/heat-operator/blob/main/pkg/heatapi/deployment.go#L152-L155

passwordSelectors have defaults defined in all layers so we don't have to define it in sub CRs. Probably it makes more sense to remove the current incomplete definitions and ensure we use the whole default map.

bshephar (approved):

PR #136, opened 2023-06-02
Refactor type definitions using templates

kajinamit:
... to reduce duplicated logics. This also allows us to hie a few spec fields in sub CRs which are not configurable in Heat CR.


api/v1beta1/common_types.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Maximum=32
+	// +kubebuilder:validation:Minimum=0
+	// Replicas -
+	Replicas int32 `json:"replicas"`
+
+	// +kubebuilder:validation:Optional
+	// NodeSelector to target subset of worker nodes for running the API service
```
bshephar, line 70: We also use this `HeatServiceTemplate` for HeatEngine, so it's not just API services. But it's a small nit

bshephar (approved):
Really nice for minimising duplication. Makes it much easier to maintain consistency when we don't need to update it for each controller type.

PR #138, opened 2023-06-02
Display status in CLI

abays:
Allows `Ready` condition status to be display at the CLI via `oc get heat`, `oc get heatapi`, etc, similar to what we do for other operators.


bshephar (approved):
Nice, sounds good.

PR #141, opened 2023-06-05
Use heat-engine image to run db sync

kajinamit:
The heat-manage command should be executed at a node with heat-engine, instead of heat-api.

### DIFF
--- a/pkg/heat/dbsync.go
+++ b/pkg/heat/dbsync.go
@@ -67,7 +67,7 @@ func DBSyncJob(
 								"/bin/bash",
 							},
 							Args:  args,
-							Image: instance.Spec.HeatAPI.ContainerImage,
+							Image: instance.Spec.HeatEngine.ContainerImage,
 							SecurityContext: &corev1.SecurityContext{
 								RunAsUser: &runAsUser,
 							},


bshephar (commented):
Might want to fix up the typo in the commit message here? I know what you're trying to say, but before we commit it to the repo, might be best to clear up the wording?

kajinamit:
> Might want to fix up the typo in the commit message here? I know what you're trying to say, but before we commit it to the repo, might be best to clear up the wording?

Yes ! and done.

Just for record, I'm wondering if we want to move database logics to HeatEngine because only heat-engine requires access to db. That allows us to drop a few more specs from top-level and might simplify the Heat CR definition.

bshephar:
> > Might want to fix up the typo in the commit message here? I know what you're trying to say, but before we commit it to the repo, might be best to clear up the wording?
> 
> Yes ! and done.
> 
> Just for record, I'm wondering if we want to move database logics to HeatEngine because only heat-engine requires access to db. That allows us to drop a few more specs from top-level and might simplify the Heat CR definition.

That sounds good. I thought I checked this with my tripleo deployment and it was the API container being used. But maybe I was wrong. In any case, that sounds like a good plan.

bshephar (approved):

kajinamit:
> 

For reference... dbsync is associated with heat-engine service here.

https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/heat/heat-engine-container-puppet.yaml#L260

and the container is defined with heat-engine image.

PR #144, opened 2023-06-06
Fix selector in affinity policy

kajinamit:
Currently we use the same name in the affinity policies to distribute pods, but this is wrong and we don't have to allocate pod for different components in different nodes. This fixes the selector and ensures we use the different name for each component.

### DIFF
--- a/controllers/heatapi_controller.go
+++ b/controllers/heatapi_controller.go
@@ -497,8 +497,7 @@ func (r *HeatAPIReconciler) reconcileNormal(ctx context.Context, instance *heatv
 	//
 
 	serviceLabels := map[string]string{
-		common.AppSelector:     heat.ServiceName,
-		heat.ComponentSelector: heat.APIComponent,
+		common.AppSelector: fmt.Sprintf("%s-%s", heat.ServiceName, heat.APIComponent),
 	}
 
 	// Handle service init
--- a/pkg/heatcfnapi/deployment.go
+++ b/pkg/heatcfnapi/deployment.go
@@ -16,6 +16,8 @@ limitations under the License.
 package heatcfnapi
 
 import (
+	"fmt"
+
 	heatv1beta1 "github.com/openstack-k8s-operators/heat-operator/api/v1beta1"
 	heat "github.com/openstack-k8s-operators/heat-operator/pkg/heat"
 	common "github.com/openstack-k8s-operators/lib-common/modules/common"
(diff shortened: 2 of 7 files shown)


controllers/heatapi_controller.go
```diff
@@ -497,8 +497,7 @@ func (r *HeatAPIReconciler) reconcileNormal(ctx context.Context, instance *heatv
 	//
 
 	serviceLabels := map[string]string{
-		common.AppSelector:     heat.ServiceName,
-		heat.ComponentSelector: heat.APIComponent,
+		common.AppSelector: fmt.Sprintf("%s-%s", heat.ServiceName, heat.APIComponent),
```
bshephar, line 500: It's quite handy to be able to do this though:
```
❯ oc get po -l service=heat
NAME                           READY   STATUS    RESTARTS   AGE
heat-api-7f775b64d-gg8gl       1/1     Running   0          10d
heat-cfnapi-5c4c8b96f7-gg6x4   1/1     Running   0          10d
heat-engine-668cd74594-k9dgk   1/1     Running   0          9d
```

By implementing this, we would no longer be able to group pods under the broader `service=heat` label and instead only be able to look up replicas of `heat-api`.

pkg/heatcfnapi/deployment.go
```diff
@@ -136,7 +139,7 @@ func Deployment(
 	deployment.Spec.Template.Spec.Affinity = affinity.DistributePods(
 		common.AppSelector,
 		[]string{
-			heat.ServiceName,
+			labels[common.AppSelector],
```
bshephar, line 142: I agree that we had a problem with all pods receiving the `api` component label. That has been fixed by the other change. But maybe we should still use two labels to preserve the common `service` grouping of pods?

bshephar: This ultimately gives the user more flexibility when trying to filter the pods their after:
```
❯ oc get po -l service=heat,component=engine
NAME                           READY   STATUS    RESTARTS   AGE
heat-engine-7fb9fb9c4d-xhf6t   1/1     Running   0          20s
```

Which would be more important in scenarios where they have multiple replicas of a deployment like `heat-engine` for example:

```
❯ oc get po -l service=heat,component=engine
NAME                           READY   STATUS     RESTARTS   AGE
heat-engine-7fb9fb9c4d-68jcr   0/1     Init:0/1   0          1s
heat-engine-7fb9fb9c4d-6p9tz   0/1     Init:0/1   0          1s
heat-engine-7fb9fb9c4d-kglc8   0/1     Init:0/1   0          1s
heat-engine-7fb9fb9c4d-xhf6t   1/1     Running    0          109s
heat-engine-7fb9fb9c4d-z66ct   0/1     Init:0/1   0          1s
```

PR #151, opened 2023-06-15
Change heat kuttl tests to run in arbitrary namespace

vhariria:


bshephar:
Looks like we haven't added kuttl tests to OpenShift ci jobs yet for Heat. So this change isn't being tested here.

bshephar (approved):
Seems fine to me and appears to be what all of the other operators are doing. We should probably get the PR up for including kuttl tests in CI jobs on this repo.

PR #156, opened 2023-06-26
Prohibit updating spec.databaseInstance

kajinamit:
Updating spec.databaseInstance does not migrate existing data to a new instance and can break resource access unexpectedly. This prohibits such update.

### DIFF
--- a/tests/functional/suite_test.go
+++ b/tests/functional/suite_test.go
@@ -90,6 +90,13 @@ var _ = BeforeSuite(func() {
 			routev1CRDs,
 		},
 		ErrorIfCRDPathMissing: true,
+		WebhookInstallOptions: envtest.WebhookInstallOptions{
+			Paths: []string{filepath.Join("..", "..", "config", "webhook")},
+			// NOTE(gibi): if localhost is resolved to ::1 (ipv6) then starting
+			// the webhook fails as it try to parse the address as ipv4 and
+			// failing on the colons in ::1
+			LocalServingHost: "127.0.0.1",
+		},
 	}
 
 	// cfg is defined in this file globally.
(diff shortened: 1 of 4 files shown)


tests/functional/suite_test.go
```diff
@@ -120,13 +127,27 @@ var _ = BeforeSuite(func() {
...
+		CertDir:            webhookInstallOptions.LocalServingCertDir,
+		LeaderElection:     false,
 	})
 	Expect(err).ToNot(HaveOccurred())
 
 	kclient, err := kubernetes.NewForConfig(cfg)
 	Expect(err).ToNot(HaveOccurred(), "failed to create kclient")
+
+	err = (&heat.Heat{}).SetupWebhookWithManager(k8sManager)
+	Expect(err).NotTo(HaveOccurred())
```
abays, line 149: I know this is testing a validation webhook instead of a defaulting one, but since we're adding the webhook server to the functional test, should we perhaps fully initialize everything required for the webhooks to function properly (in case someone later adds tests for the defaulting webhooks)?  I'm thinking specifically of this [1].

[1] https://github.com/openstack-k8s-operators/heat-operator/blob/a7985c1c308b3fed34dd36358d55864f978d5be2/main.go#L159-L160

stuggi:
https://github.com/openstack-k8s-operators/placement-operator/pull/17#issuecomment-1607070059

tests/functional/suite_test.go
kajinamit: That makes sense. Because this work is now pending, I've created PR to add tests for the defaulting webhooks.

https://github.com/openstack-k8s-operators/heat-operator/pull/179

PR #168, opened 2023-07-03
Drop functions already implemented in test helper

kajinamit:
Some functions we implement in our own test helpers are already implemented in the common test helper module in lib-common. Let's use the common implementations to reduce duplicated implementations.


bshephar (approved):
Yeah nice. Good to not duplicate all of this code.

PR #169, opened 2023-07-03
Update docs to represent current state

bshephar:
Some of the interfaces have been changed since this documentation was last updated. This change brings the documentation into consistency with the current state of the operator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,38 +23,27 @@ spec:
   rabbitMqClusterName: rabbitmq
   debug:
     dbSync: false
-    service: false
   heatAPI:
     containerImage: "quay.io/podified-antelope-centos9/openstack-heat-api:current-podified"
     customServiceConfig: "# add your customization here"
-    databaseUser: "heat"
     debug:
-      dbSync: false
       service: false
-    passwordSelectors:
-      service: HeatPassword
-      database: HeatDatabasePassword
     replicas: 1
     resources: {}
-    secret: "osp-secret"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,50 +93,43 @@ To enable the Heat service, we simply need to set the Heat service to enabled in
...
+    dbSync: false
+  heatAPI:
+    containerImage: quay.io/podified-antelope-centos9/openstack-heat-api:current-podified
+    customServiceConfig: '# add your customization here'
+    debug:
+      service: false
+    replicas: 1
+    resources: {}
+  heatCfnAPI:
+    containerImage: ""
```
kajinamit, line 112: we can probably omit containerImage, customServiceConfig and debug because these are just defining the default values and redundant.

README.md
```diff
@@ -23,38 +23,27 @@ spec:
   rabbitMqClusterName: rabbitmq
   debug:
     dbSync: false
```
kajinamit, line 25: we have to remove `spec.debug.service` in L64 (after edit)

kajinamit (changes requested):
Change looks good but let me leave a few comments.

Also, do you mind updating the pod list example in L80 because it does not contain heat-api-cfn pod ?

kajinamit (approved):
/lgtm

I could not use inline comment for unchanged line so my comment about pods listed in example was not very clear. I'm ok with fixing it in follow-up so I'll merge this now

PR #170, opened 2023-07-03
Add capability to set up notification url

kajinamit:
This introduces the capability to set up RabbitMQ transport url for notifications.

```
[oslo_messaging_notifications]
transport_url=rabbit://<user>:<password>@<host>:<port>/<vhost>
```

Because the notifications should be accessible by external consumers such as ceilometer, this change allows users to use a different RabbitMQ cluster for service transports and notifications.

Note that this change does not modify notification driver. The `[oslo_messaging_notifications] driver` option should be additionally configured by users to enable notifications over RabbitMQ.

### DIFF
--- a/api/v1beta1/conditions.go
+++ b/api/v1beta1/conditions.go
@@ -31,6 +31,9 @@ const (
 
 	// HeatStackDomainReadyCondition ...
 	HeatStackDomainReadyCondition condition.Type = "HeatStackDomainReady"
+
+	// HeatRabbitMqNotificationURLReadyCondition Status=True condition which indicates if the RabbitMQ TransportURL is ready
+	HeatRabbitMqNotificationURLReadyCondition condition.Type = "HeatRabbitMqNotificationURLReady"
 )
 
 // Common Messages used by API objects.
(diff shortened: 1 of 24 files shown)


api/v1beta1/conditions.go
```diff
@@ -60,15 +60,19 @@ const (
 )
 
 const (
-	// HeatRabbitMqTransportURLReadyCondition Status=True condition which indicates if the RabbitMQ TransportURLUrl is ready
+	// HeatRabbitMqTransportURLReadyCondition Status=True condition which indicates if the RabbitMQ TransportURL is ready
 	HeatRabbitMqTransportURLReadyCondition condition.Type = "HeatRabbitMqTransportURLReady"
+
+	// HeatRabbitMqNotificationURLReadyCondition Status=True condition which indicates if the RabbitMQ TransportURL is ready
+	HeatRabbitMqNotificationURLReadyCondition condition.Type = "HeatRabbitMqNotificationURLReady"
```
kajinamit, line 67: `NotificationTransport` is more consistent with the actual parameter name, but that looks too long so I decided to use simple `Notification` here(and the other namings).

PR #172, opened 2023-07-03
HeatAPI: Fix ExposeServiceReadyCondition

kajinamit:
This condition was not updated properly in reconciler, which results in keeping the overall condition to Unknown.

### DIFF
--- a/controllers/heatapi_controller.go
+++ b/controllers/heatapi_controller.go
@@ -329,6 +329,8 @@ func (r *HeatAPIReconciler) reconcileInit(
 		return ctrlResult, nil
 	}
 
+	instance.Status.Conditions.MarkTrue(condition.ExposeServiceReadyCondition, condition.InputReadyMessage)
+
 	// expose service - end
 
 	//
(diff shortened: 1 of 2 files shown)


bshephar (approved):

controllers/heatapi_controller.go
```diff
@@ -329,6 +329,8 @@ func (r *HeatAPIReconciler) reconcileInit(
 		return ctrlResult, nil
 	}
 
+	instance.Status.Conditions.MarkTrue(condition.ExposeServiceReadyCondition, condition.InputReadyMessage)
```
abays, line 332: Is the `condition.InputReadyMessage` right?  Should it maybe be `condition.ExposeServiceReadyMessage`?

kajinamit: Ohh, I copied it from heatcfnapi but that contains the same problem. I'll fix it soon.

kajinamit: Ignore my comment above. The implementation in cfnapi looks correct. I've submitted https://github.com/openstack-k8s-operators/heat-operator/pull/173/files .

PR #173, opened 2023-07-03
Fix wrong message used for ExposeServiceReadyCondition

kajinamit:
We should use the correct message, not the one for InputReadyCondition.


abays (approved):
/lgtm

PR #175, opened 2023-07-03
Test step-by-step condition transition in env tests

kajinamit:
This updates env tests so that we validate every single resource step in operator and assert conditions.

### DIFF
--- a/tests/functional/heat_controller_test.go
+++ b/tests/functional/heat_controller_test.go
@@ -27,12 +27,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	heatv1 "github.com/openstack-k8s-operators/heat-operator/api/v1beta1"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 )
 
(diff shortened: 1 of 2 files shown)


tests/functional/heat_controller_test.go
```diff
@@ -188,19 +290,107 @@ var _ = Describe("Heat controller", func() {
...
+		BeforeEach(func() {
+			DeferCleanup(th.DeleteInstance, CreateHeat(heatName, GetDefaultHeatSpec()))
+			DeferCleanup(
+				k8sClient.Delete, ctx, CreateHeatSecret(namespace, SecretName))
+			rmqSecret = &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "rabbitmq-secret",
+					Namespace: namespace,
+				},
+			}
```
bshephar, line 306: We do this quite a bit now. Maybe it's worth adding a function to `base_test.go` to save us reconstructing the `corev1.Secret{}` each and every time.

But it's a cosmetic thing, we can look at it later.

bshephar (approved):
Nice Takashi. Much more comprehensive, and hopefully this will help us catch any issues with setting status conditions that could potentially get stuck moving forward.

PR #177, opened 2023-07-04
Add new condition for stack domain creation

kajinamit:
This introduces the new HeatStackDomainCondition so that users can observe status of heat stack doamin creation process. Currently failures at that step are not exposed via status and users have to check pod log to find out why the creation is not progressing.

### DIFF
--- a/tests/functional/heat_controller_test.go
+++ b/tests/functional/heat_controller_test.go
@@ -88,48 +88,24 @@ var _ = Describe("Heat controller", func() {
 				condition.InputReadyCondition,
 				corev1.ConditionFalse,
 			)
-			th.ExpectCondition(
-				heatName,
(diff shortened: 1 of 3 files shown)


tests/functional/heat_controller_test.go
```diff
@@ -88,48 +88,24 @@ var _ = Describe("Heat controller", func() {
 				condition.InputReadyCondition,
 				corev1.ConditionFalse,
 			)
-			th.ExpectCondition(
-				heatName,
-				ConditionGetterFunc(HeatConditionGetter),
+
+			for _, cond := range []condition.Type{
```
bshephar, line 92: Much nicer.

bshephar (approved):
Sure, sounds good to me.

PR #178, opened 2023-07-04
Ensure TransportURL Secret exists in sub CRs

kajinamit:
Missing TransportURL Secret results in failure in the later phase and appears as a deployment failure. To give more clear error, this introduces the early validation of TransportURL Secret.

### DIFF
--- a/controllers/heatapi_controller.go
+++ b/controllers/heatapi_controller.go
@@ -428,9 +428,32 @@ func (r *HeatAPIReconciler) reconcileNormal(ctx context.Context, instance *heatv
 	// run check OpenStack secret - end
 
 	//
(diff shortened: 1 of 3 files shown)


bshephar (commented):
Are you saying that the secret existed during the first Heat controller reconcile, but then was deleted before we reconciled the sub resources?

kajinamit:
Yes that is the potential scenario caught by this change.

Though the main aim of this change is catching misbehavior of the other controllers(heat controller or even transporturl controller) properly. I've checked the other operators such as cinder/manila/ironic and noticed sub cr controllers have this validation so I think it's worth adding to our operator.

bshephar (approved):

PR #180, opened 2023-07-04
Drop invalid logic to ensure ConfigMap is not created

kajinamit:
The current logic is invalid, because ListConfigMaps returns a list of ConfigMaps in the given namespace, instead of the ones with the given name.

This drops that invalid logic until we get a proper helper function to ensure a ConfigMap does not exist.


kajinamit:
Hmm. probably we can add a helper function like AssertConfigMapDoesNotExist. let me look into that.

PR #183, opened 2023-07-06
Ensure webhook is ready before starting functional tests

kajinamit:
... to avoid executing tests with webhook not fully initialized.


bshephar (approved):
Very nice solution.

PR #186, opened 2023-07-10
Add back namespace name for kuttl config file

kajinamit:
... following the current usage in other operators (eg. keystone).


abays (approved):
/lgtm

PR #188, opened 2023-07-11
Test scaling in kuttl tests

kajinamit:
This extend the existing kuttl tests and validate scale up/down of individual services.


config/samples/heat_v1beta1_heat.yaml
```diff
@@ -3,27 +3,9 @@ kind: Heat
...
-    resources: {}
-  heatEngine:
-    customServiceConfig: "# add your customization here"
-    debug:
-      service: false
-    resources: {}
-  preserveJobs: false
+  heatAPI: {}
+  heatCfnAPI: {}
+  heatEngine: {}
```
bshephar, line 10: I have kind of been looking at the sample file as something that a user could reference to configure our operator. I can understand in the context of this change why you're doing this, but it is removing all references to things that can be configured.

I guess we need to discuss the intent behind this sample file and understand how we want it to be used. But potentially an alternative would be to copy this changed file into the tests/kuttl directory for use with these tests.

kajinamit: while we can discuss how we maintain samples and test coverage, I'm wondering if we can unblock this PR to get a better test coverage. I can restore a few specific options if you think these are very useful, but I don't think all of the existing options such as databaseUser or serviceUser are very useful even of user references. Probably customServiceConfig ?

bshephar (approved):

PR #190, opened 2023-07-11
Run tests concurrently

kajinamit:
This is a follow-up of the previous change which replaces go test by ginkgo and fixes the missing PROC_CMD variable so that all tests are executed concurrently.


bshephar (approved):
Oh nice one!

PR #192, opened 2023-07-11
Update test helper references after reorg

gibizer:
Depends-On: openstack-k8s-operators/lib-common#301


stuggi:
https://github.com/openstack-k8s-operators/heat-operator/blob/main/tests/functional/suite_test.go#L33-L37 should also be updated I think

bshephar (approved):

stuggi (approved):
/lgtm

PR #200, opened 2023-07-18
Fix override of CustomServiceConfig

kajinamit:
Currently CustomServiceConfig at top-layer and each sub CR layer are independent and users have to configure all fields to add the same config settings to all services.

This fixes that behavior and ensures the top-level CustomServiceConfig is used even when the sub-level CustomServiceConfig is used.

### DIFF
--- a/pkg/heat/const.go
+++ b/pkg/heat/const.go
@@ -30,6 +30,12 @@ const (
 	StackDomainName = "heat_stack"
 	// DatabaseName -
 	DatabaseName = "heat"
+	// DefaultsConfigFileName -
+	DefaultsConfigFileName = "00-config.conf"
+	// CustomConfigFileName -
+	CustomConfigFileName = "01-config.conf"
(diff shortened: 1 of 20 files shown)


pkg/heat/const.go
```diff
@@ -30,6 +30,12 @@ const (
 	StackDomainName = "heat_stack"
 	// DatabaseName -
 	DatabaseName = "heat"
+	// DefaultsConfigFileName -
+	DefaultsConfigFileName = "00-config.conf"
+	// CustomConfigFileName -
+	CustomConfigFileName = "01-config.conf"
+	// CustomServiceConfigFileName -
+	CustomServiceConfigFileName = "02-config.conf"
```
bshephar, line 38: How do we plan to use these?

kajinamit:
Umm. I noticed this merging is handled by crudini. Probably I have to change focus of this change.

I'll recreate PR with the proper title.

PR #203, opened 2023-07-19
Configure renovate to pin k8s and controller-runtime for dev preview and pin ocp

stuggi:
- pin k8s and controller runtime to be in sync with lib-common and configure renovate to keep the pin
- pin ocp api version to a hash in sync with lib-common instead of the nonexistent v3.9.0+incompatible tag

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -87,3 +87,7 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+// mschuppert: map to latest commit from release-4.13 tag
+// must consistent within modules and service operators
+replace github.com/openshift/api => github.com/openshift/api v0.0.0-20230414143018-3367bc7e6ac7 //allow-merging
--- a/renovate.json
+++ b/renovate.json
@@ -27,14 +27,16 @@
     },
     {
       "groupName": "k8s.io",
-      "matchPackagePrefixes": [
-        "k8s.io",
-        "sigs.k8s.io"
-      ],
-      "schedule": [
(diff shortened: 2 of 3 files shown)


gibizer (approved):

go.mod
```diff
@@ -87,3 +87,7 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+// mschuppert: map to latest commit from release-4.13 tag
+// must consistent within modules and service operators
+replace github.com/openshift/api => github.com/openshift/api v0.0.0-20230414143018-3367bc7e6ac7 //allow-merging
```
bshephar, line 93: Yeah, I think this is probably the easiest way to manage this.

renovate.json
```diff
@@ -27,14 +27,16 @@
...
+      "matchPackagePatterns": ["^k8s.io"],
+      "excludePackagePatterns": ["^k8s.io/kube-openapi"],
+      "allowedVersions": "< 0.27.0",
+      "enabled": true
+    },
+    {
+      "groupName": "sigs.k8s.io/controller-runtime",
+      "matchPackagePatterns": ["^sigs.k8s.io/controller-runtime"],
+      "allowedVersions": "< 0.15.0",
+      "enabled": true
```
bshephar, line 39: I guess this will also catch things like gnostic and go-openapi that are indirect imports, so this should be safe.

bshephar (approved):
Looks good to me. Are we looking to merge this now?

kajinamit:
/retest

Requesting bear token: invalid status code from registry 500 (Internal Server Error)

PR #213, opened 2023-07-27
Enable debug log for heat services

kajinamit:
... so that we can check behavior of heat services more closely in CI. This also removes the default value for customServiceConfig following the other operators.


abays:
This file also needs to be updated: https://github.com/openstack-k8s-operators/heat-operator/blob/main/tests/kuttl/common/assert-sample-deployment.yaml

bshephar (approved):

PR #218, opened 2023-08-01
Refactor creation of heat stack domain

kajinamit:
We have multiple functions but each of these is tiny. Squashing these into a single function saves us from passing multiple instances to each layer.

Also this removes redundant lookup of the global secret.

### DIFF
--- a/controllers/heat_controller.go
+++ b/controllers/heat_controller.go
@@ -13,6 +13,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


controllers/heat_controller.go
```diff
@@ -898,7 +899,15 @@ func (r *HeatReconciler) ensureStackDomain(
 	ctx context.Context,
 	helper *helper.Helper,
 	instance *heatv1beta1.Heat,
+	secret *corev1.Secret,
 ) (ctrl.Result, error) {
+
+	val, ok := secret.Data[instance.Spec.PasswordSelectors.Service]
+	if !ok {
+		return ctrl.Result{}, fmt.Errorf("%s not foind in secret %s", instance.Spec.PasswordSelectors.Service, instance.Spec.Secret)
```
bshephar, line 907: /foind/found/

kajinamit: me~~  fixed in the latest version.

bshephar (approved):

PR #221, opened 2023-08-03
Remove redundant defaults of PasswordSelector

kajinamit:
We define defaults in two layers but this is redundant. This change attempts to define defaults in the single layer instead.

### DIFF
--- a/api/bases/heat.openstack.org_heatapis.yaml
+++ b/api/bases/heat.openstack.org_heatapis.yaml
@@ -88,10 +88,6 @@ spec:
                   the service
                 type: object
               passwordSelectors:
-                default:
-                  authEncryptionKey: HeatAuthEncryptionKey
-                  database: HeatDatabasePassword
-                  service: HeatPassword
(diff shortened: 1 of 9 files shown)


bshephar (approved):

bshephar:
Looks like it's just complaining about:
```
 # github.com/openstack-k8s-operators/heat-operator/api/v1beta1
api/v1beta1/common_types.go:92:2: C003: Field 'Database' has both a 'Optional' kubebuilder marker with a default value and an 'omitempty' tag. Either remove the default value or remove 'omitempty'
api/v1beta1/common_types.go:96:2: C003: Field 'Service' has both a 'Optional' kubebuilder marker with a default value and an 'omitempty' tag. Either remove the default value or remove 'omitempty'
api/v1beta1/common_types.go:100:2: C003: Field 'AuthEncryptionKey' has both a 'Optional' kubebuilder marker with a default value and an 'omitempty' tag. Either remove the default value or remove 'omitempty' 
```

PR #222, opened 2023-08-03
Simplify if else statements

bshephar:
If we're returning from a if statement, we really don't need an else. These can be simplified and made more legible by simply removing the else and having a seperate if block.

This is really just a quality of life patch. IMO, the use of `else` can often make code a bit harder to follow and read. 

If this is just something we're doing for consistency with the wider org's sake, I'm happy to raise it to the wider group for broader perspective. But at least here, I think it makes things a little nicer to read and follow.

### DIFF
--- a/controllers/heat_controller.go
+++ b/controllers/heat_controller.go
@@ -294,10 +294,8 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 		},
 	}
 	rbacResult, err := common_rbac.ReconcileRbac(ctx, helper, instance, rbacRules)
-	if err != nil {
+	if err != nil || (rbacResult != ctrl.Result{}) {
 		return rbacResult, err
-	} else if (rbacResult != ctrl.Result{}) {
-		return rbacResult, nil
 	}
 
 	// ConfigMap
@@ -437,26 +435,20 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 
 	// Handle service init
 	ctrlResult, err := r.reconcileInit(ctx, instance, helper, serviceLabels)
-	if err != nil {
+	if err != nil || (ctrlResult != ctrl.Result{}) {
 		return ctrlResult, err
-	} else if (ctrlResult != ctrl.Result{}) {
-		return ctrlResult, nil
 	}
 
 	// Handle service update
 	ctrlResult, err = r.reconcileUpdate(ctx, instance, helper)
-	if err != nil {
+	if err != nil || (ctrlResult != ctrl.Result{}) {
 		return ctrlResult, err
-	} else if (ctrlResult != ctrl.Result{}) {
-		return ctrlResult, nil
 	}
 
 	// Handle service upgrade
 	ctrlResult, err = r.reconcileUpgrade(ctx, instance, helper)
-	if err != nil {
+	if err != nil || (ctrlResult != ctrl.Result{}) {
 		return ctrlResult, err
-	} else if (ctrlResult != ctrl.Result{}) {
-		return ctrlResult, nil
 	}
 
 	//
@@ -473,7 +465,8 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 			heatv1beta1.HeatStackDomainReadyErrorMessage,
 			err.Error()))
 		return ctrl.Result{}, err
-	} else if (ctrlResult != ctrl.Result{}) {
+	}
+	if (ctrlResult != ctrl.Result{}) {
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			heatv1beta1.HeatStackDomainReadyCondition,
 			condition.RequestedReason,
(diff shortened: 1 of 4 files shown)


controllers/heat_controller.go
```diff
@@ -296,7 +296,8 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 	rbacResult, err := common_rbac.ReconcileRbac(ctx, helper, instance, rbacRules)
 	if err != nil {
```
kajinamit, line 297: wondering if we can just do

```
if err != nil || rbacResult != ctrl.Result{} {
    return rbacResult, err
}
```

bshephar: Very true, yeah, let's do that instead.

kajinamit:
/test functional

Is this a consistent failure ?

bshephar:
Hmm yeah, this is reproducing locally as well actually.

bshephar:
Interesting, seems to be this one:
https://github.com/openstack-k8s-operators/heat-operator/compare/af1729757e111269defb133988a2474d9a4f3b7d..1dc91656281bd84f234487ca4a95c6610afeb5ab

kajinamit:
> Interesting, seems to be this one: https://github.com/openstack-k8s-operators/heat-operator/compare/af1729757e111269defb133988a2474d9a4f3b7d..1dc91656281bd84f234487ca4a95c6610afeb5ab
> 
> ```
> 	//
> 	// get admin authentication OpenStack
> 	//
> 	os, ctrlResult, err := keystonev1.GetAdminServiceClient(
> 		ctx,
> 		helper,
> 		keystoneAPI,
> 	)
> 	fmt.Printf("bshephar-debug: err: %+v\n ctrlResult: %+v \n comp ctrl.Result: %+v\n", err, ctrlResult, ctrl.Result{})
> 	if err != nil || (ctrlResult != ctrl.Result{}) {
> 		return ctrl.Result{}, err
> 	}
> ```
> 
> There's a diff in ctrlResult vs ctrl.Result{}:
> 
> ```
>   bshephar-debug: err: <nil>
>    ctrlResult: {Requeue:false RequeueAfter:10s} 
>    comp ctrl.Result: {Requeue:false RequeueAfter:0s}
>   bshephar-debug: err: <nil>
>    ctrlResult: {Requeue:false RequeueAfter:10s} 
>    comp ctrl.Result: {Requeue:false RequeueAfter:0s}
>   bshephar-debug: err: <nil>
>    ctrlResult: {Requeue:false RequeueAfter:10s} 
>    comp ctrl.Result: {Requeue:false RequeueAfter:0s}
>   bshephar-debug: err: <nil>
>    ctrlResult: {Requeue:false RequeueAfter:10s} 
>    comp ctrl.Result: {Requeue:false RequeueAfter:0s}
>   bshephar-debug: err: <nil>
>    ctrlResult: {Requeue:false RequeueAfter:10s} 
>    comp ctrl.Result: {Requeue:false RequeueAfter:0s}
>   bshephar-debug: err: <nil>
>    ctrlResult: {Requeue:false RequeueAfter:10s} 
>    comp ctrl.Result: {Requeue:false RequeueAfter:0s}
>   bshephar-debug: err: <nil>
>    ctrlResult: {Requeue:false RequeueAfter:10s} 
>    comp ctrl.Result: {Requeue:false RequeueAfter:0s}
> ```
> 
> But not sure just yet as to why it would make a difference in the consolidated if this or this, vs the two seperate if statements.
> 
> It's exactly the same except it doesn't go on for as long when I do the same thing with seperate if statements:
> 
> ```
>   bshephar-debug: err: <nil>
>    ctrlResult: {Requeue:false RequeueAfter:10s} 
>    comp ctrl.Result: {Requeue:false RequeueAfter:0s}
>   bshephar-debug: err: <nil>
>    ctrlResult: {Requeue:false RequeueAfter:10s} 
>    comp ctrl.Result: {Requeue:false RequeueAfter:0s}
>   bshephar-debug: err: <nil>
>    ctrlResult: {Requeue:false RequeueAfter:10s} 
>    comp ctrl.Result: {Requeue:false RequeueAfter:0s}
> ```
> 
> Maybe some timing thing?

It looks you returned `ctrl.Result{}` instead of `ctrlResult` and I guess that is the problem.

controllers/heatcfnapi_controller.go
```diff
@@ -336,7 +337,7 @@ func (r *HeatCfnAPIReconciler) reconcileInit(
 
 		ksSvcObj := keystonev1.NewKeystoneService(ksSvcSpec, instance.Namespace, serviceLabels, time.Duration(10)*time.Second)
 		ctrlResult, err = ksSvcObj.CreateOrPatch(ctx, helper)
-		if err != nil {
+		if err != nil || (ctrlResult != ctrl.Result{}) {
```
kajinamit, line 340: I thought we'd need this logic kepy after condition mirror but it turned out this method returns non-emptry ctrlResult only when KeystoneService is not found... so skipping the condition mirror makes more sense.

PR #237, opened 2023-09-05
Add .prow_ci.env

dprince:
This is used to enable USE_IMAGE_DIGESTS for prow builds

### DIFF
--- a/.prow_ci.env
+++ b/.prow_ci.env
@@ -1,0 +1,1 @@
+export USE_IMAGE_DIGESTS=true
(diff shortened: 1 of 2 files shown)


kajinamit:
`case.go:366: resource HeatEngine:heat-kuttl-tests/heat-engine: .spec.containerImage: value mismatch, expected: quay.io/podified-antelope-centos9/openstack-heat-engine:current-podified != actual: quay.io/podified-antelope-centos9/openstack-heat-engine@sha256:c8295d1f33c1308b49ebb88c69914c510d376392ac27c943cb0b4866d23cd11f`

It seems we have to update kuttl tests ?

PR #252, opened 2023-10-03
Block changing DatabaseInstance

bshephar:
We currently have no mechanism to validate the new database will match the existing database. This could ultimatly result in user objects being orphaned and unrecoverable. We may need to implement a bypass for this webhook in cases where we are doing adoptions and have independently verified the contents of the databases. But without any kind of validation, we should block users from inadvertently causing potentially unrecoverable errors.

### DIFF
--- a/api/v1beta1/heat_webhook.go
+++ b/api/v1beta1/heat_webhook.go
@@ -23,7 +23,12 @@ limitations under the License.
 package v1beta1
 
 import (
+	"fmt"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
@@ -94,8 +99,30 @@ func (r *Heat) ValidateCreate() error {
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *Heat) ValidateUpdate(old runtime.Object) error {
 	heatlog.Info("validate update", "name", r.Name)
+	oldHeat, ok := old.(*Heat)
+	if !ok {
+		return apierrors.NewInternalError(
+			fmt.Errorf("Expected a Heatv1 object, but got %T", oldHeat))
+	}
 
-	// TODO(user): fill in your validation logic upon object update.
+	var errors field.ErrorList
+	// We currently have no logic in place to perform database migrations. Changing databases
+	// would render all of the existing stacks unmanageable. We should block changes to the
+	// databaseInstance to protect existing workloads.
+	if r.Spec.DatabaseInstance != oldHeat.Spec.DatabaseInstance {
+		errors = append(errors, field.Forbidden(
+			field.NewPath("spec.databaseInstance"),
+			"Changing the DatabaseInstance is not supported for existing deployments"))
+	}
+
+	if errors != nil {
+		return apierrors.NewInvalid(
+			schema.GroupKind{Group: "heat.openstack.org", Kind: "Heat"},
+			r.Name,
+			errors,
+		)
+	}
+
 	return nil
 }
 
--- a/tests/functional/heat_webhook_test.go
+++ b/tests/functional/heat_webhook_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package functional_test
 
 import (
+	"fmt"
 	"os"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -86,6 +87,21 @@ var _ = Describe("Heat Webhook", func() {
 			Expect(Heat.Spec.HeatEngine.ContainerImage).Should(Equal(
 				"engine-container-image",
 			))
+		})
+	})
+
+	When("The DatabaseInstance is changed for existing deployments", func() {
+		BeforeEach(func() {
+			DeferCleanup(th.DeleteInstance, CreateHeat(heatName, GetDefaultHeatSpec()))
+		})
+
+		It("Should be blocked by the webhook", func() {
+			Eventually(func(g Gomega) string {
+				instance := GetHeat(heatName)
+				instance.Spec.DatabaseInstance = "new-database"
+				err := th.K8sClient.Update(th.Ctx, instance)
+				return fmt.Sprintf("%s", err)
+			}).Should(ContainSubstring("Changing the DatabaseInstance is not supported for existing deployments"))
 		})
 	})
 })


gibizer:
There was a long discussion around this topic in https://github.com/openstack-k8s-operators/placement-operator/pull/17  I haven't load all the context from that discussion. But in that PR there was suggestion to allow overriding the write-once check on the databaseInstance. The only known case for that was adoption. But I think adoption does not need databaseInstance change. We deploy the control plane with new DB servers, and the copy the data from the old deployment to the new DB server. This data migration does not require databaseInstance change on the new deployment. So right now I'm on the side of not providing an override to the write-once check. We can add that later when a use case appear for that. 

(cc @stuggi @SeanMooney @kajinamit @bogdando as they commented on the original PR)

bshephar:
> There was a long discussion around this topic in [openstack-k8s-operators/placement-operator#17](https://github.com/openstack-k8s-operators/placement-operator/pull/17) I haven't load all the context from that discussion. But in that PR there was suggestion to allow overriding the write-once check on the databaseInstance. The only known case for that was adoption. But I think adoption does not need databaseInstance change. We deploy the control plane with new DB servers, and the copy the data from the old deployment to the new DB server. This data migration does not require databaseInstance change on the new deployment. So right now I'm on the side of not providing an override to the write-once check. We can add that later when a use case appear for that.
> 
> (cc @stuggi @SeanMooney @kajinamit @bogdando as they commented on the original PR)

Yeah, I don't think we need an interface to override it. We seem safe to just block the change at this point, but if anyone else had any concerns, happy to hear them.

gibizer (commented):
I'm OK with the direction

tests/functional/heat_webhook_test.go
```diff
@@ -88,4 +88,16 @@ var _ = Describe("Heat Webhook", func() {
...
+
+	When("The DatabaseInstance is changed for existing deployments", func() {
+		BeforeEach(func() {
+			DeferCleanup(th.DeleteInstance, CreateHeat(heatName, GetDefaultHeatSpec()))
+		})
+
+		It("Should be blocked by the webhook", func() {
+			instance := GetHeat(heatName)
+			instance.Spec.DatabaseInstance = "new-database"
+			Expect(th.K8sClient.Update(th.Ctx, instance)).ShouldNot(Succeed())
```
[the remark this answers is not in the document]
bshephar: Addressed this one with the latest revision.

api/v1beta1/heat_webhook.go
```diff
@@ -94,8 +96,18 @@ func (r *Heat) ValidateCreate() error {
...
+	if !ok {
+		return fmt.Errorf("Expected a Heatv1 object, but got %T", oldHeat)
+	}
+
+	// We currently have no logic in place to perform database migrations. Changing databases
+	// would render all of the existing stacks unmanageable. We should block changes to the
+	// databaseInstance to protect existing workloads.
+	if r.Spec.DatabaseInstance != oldHeat.Spec.DatabaseInstance {
+		return fmt.Errorf("Changing the DatabaseInstance is not supported for existing deployments")
+	}
```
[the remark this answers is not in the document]
bshephar: Let me know if the latest update here addresses this concern.

api/v1beta1/heat_webhook.go
```diff
@@ -94,8 +99,30 @@ func (r *Heat) ValidateCreate() error {
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *Heat) ValidateUpdate(old runtime.Object) error {
 	heatlog.Info("validate update", "name", r.Name)
+	oldHeat, ok := old.(*Heat)
+	if !ok {
+		return fmt.Errorf("Expected a Heatv1 object, but got %T", oldHeat)
```
stuggi, line 104: maybe should return a `apierrors.NewInternalError( fmt.Errorf())` 
```
NewInternalError returns an error indicating the item is invalid and cannot be processed.
```

api/v1beta1/heat_webhook.go
```diff
@@ -94,8 +99,30 @@ func (r *Heat) ValidateCreate() error {
...
+	}
+
+	var errors field.ErrorList
+	// We currently have no logic in place to perform database migrations. Changing databases
+	// would render all of the existing stacks unmanageable. We should block changes to the
+	// databaseInstance to protect existing workloads.
+	if r.Spec.DatabaseInstance != oldHeat.Spec.DatabaseInstance {
+		errors = append(errors, field.Invalid(
```
stuggi, line 112: should we use `field.Forbidden` instead since we say its a forbidden change?

SeanMooney:
> > There was a long discussion around this topic in [openstack-k8s-operators/placement-operator#17](https://github.com/openstack-k8s-operators/placement-operator/pull/17) I haven't load all the context from that discussion. But in that PR there was suggestion to allow overriding the write-once check on the databaseInstance. The only known case for that was adoption. But I think adoption does not need databaseInstance change. We deploy the control plane with new DB servers, and the copy the data from the old deployment to the new DB server. This data migration does not require databaseInstance change on the new deployment. So right now I'm on the side of not providing an override to the write-once check. We can add that later when a use case appear for that.
> > (cc @stuggi @SeanMooney @kajinamit @bogdando as they commented on the original PR)
> 
> Yeah, I don't think we need an interface to override it. We seem safe to just block the change at this point, but if anyone else had any concerns, happy to hear them.

so the original usecase for allowing changing the database instance was not for adoption.
it was for changing the topology form a single shared database to a database per service or splitting out some subset fo services to there own db.

as clouds grow and load increases it can be increasingly advantageous to move from the unified to split-out topology for performance reasons. yes that would require a manual data copy to be done first and yes you could get the same effect by deleting the nova/neutron cr and redefining it with the new db instances but I'm really not convinced that we need to block it.

Are we going to do the same for rabbitmq? if you incorrectly set the transport URL to the wrong rabbit and it talks to the incorrect cell conductor its pretty catastofic in terms of the db currption that causes are we going to also block that in the dataplane operator somehow?

if we do this it should be done for all datastores including memcache, redes and rabbit not just mysql.

bshephar:
> > > There was a long discussion around this topic in [openstack-k8s-operators/placement-operator#17](https://github.com/openstack-k8s-operators/placement-operator/pull/17) I haven't load all the context from that discussion. But in that PR there was suggestion to allow overriding the write-once check on the databaseInstance. The only known case for that was adoption. But I think adoption does not need databaseInstance change. We deploy the control plane with new DB servers, and the copy the data from the old deployment to the new DB server. This data migration does not require databaseInstance change on the new deployment. So right now I'm on the side of not providing an override to the write-once check. We can add that later when a use case appear for that.
> > > (cc @stuggi @SeanMooney @kajinamit @bogdando as they commented on the original PR)
> > 
> > 
> > Yeah, I don't think we need an interface to override it. We seem safe to just block the change at this point, but if anyone else had any concerns, happy to hear them.
> 
> so the original usecase for allowing changing the database instance was not for adoption. it was for changing the topology form a single shared database to a database per service or splitting out some subset fo services to there own db.
> 
> as clouds grow and load increases it can be increasingly advantageous to move from the unified to split-out topology for performance reasons. yes that would require a manual data copy to be done first and yes you could get the same effect by deleting the nova/neutron cr and redefining it with the new db instances but I'm really not convinced that we need to block it.
> 
> Are we going to do the same for rabbitmq? if you incorrectly set the transport URL to the wrong rabbit and it talks to the incorrect cell conductor its pretty catastofic in terms of the db currption that causes are we going to also block that in the dataplane operator somehow?
> 
> if we do this it should be done for all datastores including memcache, redes and rabbit not just mysql.

So, would you prefer to do it an also allow it to be bypassed like I initially proposed in:
https://github.com/openstack-k8s-operators/heat-operator/pull/253

I do think we need some protection against erroneous changes of this, since it will potentially be a pain to recover. And, if it _can_ be changed by users, it _will_ be changed by users. But I would definitely be open to allowing it to be bypassed under the right circumstances.

api/v1beta1/heat_webhook.go
bshephar: Sounds reasonable, changed this now.

api/v1beta1/heat_webhook.go
bshephar: Yeah, sounds good. Changed it.

dprince (approved):

dprince:
@bshephar I think this looks good. Probably want to consider adding the same to other operators as well.

PR #270, opened 2023-11-06
Use internal Keystone endpoint

bshephar:
This change updates the heat.conf file to ensure our connections to Keystone happen via the internal endpoint rather than the public one.


stuggi (approved):
/lgtm

PR #297, opened 2024-01-10
Use internal endpoints for clients by default

rabi:
We've enabled TLS for public endpoints by default[1] and heat default uses public endpoints for clients. Let's override to use internal endpoints.

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/core/v1beta1/openstackcontrolplane_types.go#L78


templates/heat/config/heat.conf
```diff
@@ -49,3 +49,7 @@ user_domain_name=Default
 project_name=service
 project_domain_name=Default
 interface=internal
+
+[clients]
+endpoint_type=internal
+insecure=true
```
stuggi, line 55: the insecure should be not needed, right? when we enable tls for internal endpoints we also add the CAs to the deployment pods

rabi: Yeah we can drop it as `insecure=false` is the default. However, insecure=false only makes sense when there are certs to be validated.

stuggi (approved):
/lgtm

PR #300, opened 2024-01-19
[tlse] tls for heat pod configuration

stuggi:
Public/Internal service cert secrets and the CA bundle secret can be passed to configure httpd virtual hosts for tls termination. The CA cert get direct mounted as the environment bundle to /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem . The service certificates like config files and copied via kolla to /etc/pki/tls/certs/%s.crt|/etc/pki/tls/private/%s.key .
Job deployments for bootstrap/cron get the CA bundle added if configured.

Also indexes the named input resources for password, CA bundle, and endpoint secrets to be able to watch them for a change and reconcile.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/428

Jira: [OSPRH-3851](https://issues.redhat.com//browse/OSPRH-3851)

Follow up: envtest for tls, when general test coverage for heat is added


bshephar (approved):
Very nice. A lot of work went into this one. We should probably have some functional tests for all of this new code too, but looks like it passes kuttl now, so that's a good start.

PR #313, opened 2024-02-14
Controller runtime v0.16.4 Go 1.20

bshephar:


go.mod
```diff
@@ -1,44 +1,44 @@
...
 	github.com/openstack-k8s-operators/mariadb-operator/api v0.3.1-0.20240208072109-4447f245e487
-	k8s.io/api v0.26.13
-	k8s.io/apimachinery v0.26.13
-	k8s.io/client-go v0.26.13
+	k8s.io/api v0.29.1
+	k8s.io/apimachinery v0.29.1
+	k8s.io/client-go v0.29.1
 	k8s.io/utils v0.0.0-20240102154912-e7106e64919e
-	sigs.k8s.io/controller-runtime v0.14.7
+	sigs.k8s.io/controller-runtime v0.17.0
```
gibizer, line 23: I this should be 0.16

bshephar: I think it's because I was just trying to downgrade controller-runtime which didn't auto downgrade the k8s.io dependencies, then mod tidy puts this back.

This seems more reliable now:
```
go1.20.3 get -u k8s.io/api@v0.28.3
```

Which also downgrades controller-runtime to 0.16.4.

PR #322, opened 2024-02-20
migrate from databaseUsername to databaseAccount and fully use MariaDBAccount

zzzeek:
This moves heat to fully use MariaDBAccount based on the dev work being done for mariadb-operator:

Lead Jira: [OSPRH-4095](https://issues.redhat.com/browse/OSPRH-4095)

1. [x] controller calls EnsureMariaDBAccount up front to make sure MariaDBAccount is present
2. [x] error code from EnsureMariaDBAccount is handled, Conditions are amended when error is returned
3. [x] controller calls NewDatabaseForAccount instead of NewDatabase
4. [x]  GetAccountAndSecret is used to retrieve account /secret to populate template
5. [x]  GetDatabaseByName() , normally used for delete finalizers, replaced with GetDatabaseByNameAndAccount
6. [x]  CreateOrPatchAll() used to patch the Database, replacing CreateOrPatchDB / CreateOrPatchDBByName
7. [x]  controller calls DeleteUnusedMariaDBAccountFinalizers when launched pods are definitely running on a new MariaDBAccount, returns error code if present
8. [x]  PasswordSelectors that refer to database are removed
9. [x]  all databaseUser replaced with databaseAccount inside of all XYZ_types.go
10. [x] all databaseUser replaced with databaseAccount inside of all `kuttl/*.yaml`
11. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all XYZ_types.go
12. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all `kuttl/*.yaml`
13. [x] MariaDBAccountSuiteTests are used in controller ginkgo tests if it has them
14. [x] Use configsecrets for database URLs; remove from job hash
13. [x] [184](https://github.com/openstack-k8s-operators/mariadb-operator/pull/184) is merged and replaces from go.mod are removed

### DIFF
--- a/controllers/heat_controller.go
+++ b/controllers/heat_controller.go
@@ -20,10 +20,10 @@ import (
 
 	"github.com/openstack-k8s-operators/lib-common/modules/common"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
-	configmap "github.com/openstack-k8s-operators/lib-common/modules/common/configmap"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/endpoint"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/job"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
 
@@ -84,7 +84,7 @@ var keystoneAPI *keystonev1.KeystoneAPI
 // +kubebuilder:rbac:groups=heat.openstack.org,resources=heatengines/finalizers,verbs=update
 // +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete;
-// +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;
+// +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;create;update;patch;delete;watch
 // +kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete;
 // +kubebuilder:rbac:groups=mariadb.openstack.org,resources=mariadbdatabases,verbs=get;list;watch;create;update;patch;delete;
@@ -334,7 +334,7 @@ func (r *HeatReconciler) reconcileDelete(ctx context.Context, instance *heatv1be
 	r.Log.Info("Reconciling Heat delete")
 
 	// remove db finalizer first
-	db, err := mariadbv1.GetDatabaseByName(ctx, helper, instance.Name)
+	db, err := mariadbv1.GetDatabaseByNameAndAccount(ctx, helper, instance.Name, instance.Spec.DatabaseAccount, instance.Namespace)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return ctrl.Result{}, err
 	}
@@ -568,6 +568,22 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 		return ctrlResult, err
 	}
 
+	// remove finalizers from previous MariaDBAccounts for which we have
+	// switched.
+	// TODO(zzzeek) - It's not clear if this is called too early here.
+	// at the moment, heat_controller_test.go doesn't seem to have fixtures
+	// I can use to simulate getting all the way to the end of a reconcile
+	// for an instance.  Basically this should be called when any pods have
+	// been restarted to run on an updated set of DB credentials, and the old
+	// ones are no longer needed.  This would allow the scenario where
+	// a new MariaDBAccount is created and an old MariaDBAccount is marked
+	// deleted at once, where the finalizer will keep the old one around until
+	// it's safe to drop.
+	err = mariadbv1.DeleteUnusedMariaDBAccountFinalizers(ctx, helper, instance.Name, instance.Spec.DatabaseAccount, instance.Namespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	//
 	// normal reconcile tasks
 	//
@@ -583,6 +599,7 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 			err.Error()))
 		return ctrl.Result{}, err
 	}
+
 	if (ctrlResult != ctrl.Result{}) {
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			heatv1beta1.HeatStackDomainReadyCondition,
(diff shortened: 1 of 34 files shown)


zzzeek:
@gibizer if you want to look at this one, with a few of these that had database username/password sent as env to all job templates, I had to change that since a rotation of username/password meant the dbsync job would run again, etc.   I moved these to use configmaps with a "databaseConnection" token containing the URL in them the way nova/neutron/keystone do, but also, I moved from "configmap" to "configsecret" (also how you can see keystone doing it already) since these contain passwords in them and the whole point of what we're doing is trying to increase username/password security.   so that has implications for the volume mounts which become "secret" mounts.    designate is another one where I did this, which i haven't updated to the latest yet.

controllers/heat_controller.go
```diff
@@ -568,6 +568,17 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 		return ctrlResult, err
 	}
 
+	// remove finalizers from previous MariaDBAccounts for which we have
+	// switched.
+	// TODO(zzzeek) - This is here because I dont have test fixtures to
```
zzzeek, line 573: this comment refers to when is it safe to delete finalizers from old mariadbaccounts, meaning, are we certain that all pods are re-deployed with a new username/password by the time we get here

controllers/heat_controller.go
```diff
@@ -930,7 +956,7 @@ func (r *HeatReconciler) generateServiceConfigMaps(
 			Labels:        cmLabels,
 		},
 	}
-	return configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
+	return secret.EnsureSecrets(ctx, h, instance, cms, envVars)
```
zzzeek, line 959: here's the switch to secrets

controllers/heatcfnapi_controller.go
```diff
@@ -810,6 +797,38 @@ func (r *HeatCfnAPIReconciler) reconcileUpgrade(ctx context.Context, instance *h
 	return ctrl.Result{}, nil
 }
 
+// getSecret - get the specified secret, and add its hash to envVars
+func (r *HeatCfnAPIReconciler) getSecret(
```
zzzeek, line 801: note this approach with getSecret is copied from designate which has these functions in each controller as well.  we can try to improve this but it would add more dev time

pkg/heat/dbsync.go
```diff
@@ -88,10 +88,8 @@ func DBSyncJob(
 	initContainerDetails := APIDetails{
 		ContainerImage:            instance.Spec.HeatAPI.ContainerImage,
 		DatabaseHost:              instance.Status.DatabaseHostname,
-		DatabaseUser:              instance.Spec.DatabaseUser,
```
zzzeek, line 91: this is what had to be removed otherwise we get new DBSyncJob everytime username changes

pkg/heat/volumes.go
```diff
@@ -28,22 +28,18 @@ func GetVolumes(name string) []corev1.Volume {
 		{
 			Name: "scripts",
 			VolumeSource: corev1.VolumeSource{
-				ConfigMap: &corev1.ConfigMapVolumeSource{
+				Secret: &corev1.SecretVolumeSource{
```
zzzeek, line 31: using all secrets, not configmaps

templates/heat/bin/init.sh
```diff
@@ -73,7 +69,6 @@ if [ -n "$AUTH_ENCRYPTION_KEY" ]; then
     crudini --set ${SVC_CFG_MERGED} DEFAULT auth_encryption_key $AUTH_ENCRYPTION_KEY
 fi
 
-crudini --set ${SVC_CFG_MERGED} database connection mysql+pymysql://${DBUSER}:${DBPASSWORD}@${DBHOST}/${DB}?read_default_file=/etc/my.cnf
```
zzzeek, line 76: and finally, the payoff for all that configmap stuff, removing this

templates/heat/config/heat.conf
```diff
@@ -13,7 +13,7 @@ enabled=true
 memcache_servers={{ .MemcachedServers }}
 
 [database]
-#connection =
+connection={{ .DatabaseConnection }}
```
zzzeek, line 16: and the URL goes here instead

controllers/heatengine_controller.go
```diff
@@ -528,6 +515,38 @@ func (r *HeatEngineReconciler) reconcileUpgrade(ctx context.Context, instance *h
 	return ctrl.Result{}, nil
 }
 
+// getSecret - get the specified secret, and add its hash to envVars
+func (r *HeatEngineReconciler) getSecret(
+	ctx context.Context,
+	h *helper.Helper,
+	instance *heatv1beta1.HeatEngine,
```
bshephar, line 522: I'm not a huge fan of this being duplicated for each controller. I believe the instance here would satisfy the `client.Object` interface, which would mean we can create a single function that isn't a method of the reconciler, where the instance passed in is of type `client.Object`. 

But we can merge this now and then address that point in a follow up PR. I assume all of the other PR's are doing the same thing as defined here. So for consistency sake, it's best to just merge and we'll look at improving the implementation later.

bshephar (approved):

controllers/heatengine_controller.go
bshephar: Ah, we get Status. That isn't a client.Object field. But we still might be able to find a common type that satisfies this requirement and enables us to make this code more generic. I'll dig into it and see if I can propose something that we can collab on.

zzzeek: I still have three more of these PRs to do, so far every single one has had some new thing I've had to learn so we'll see how barbican / manila / telemetry go

controllers/heatapi_controller.go
```diff
@@ -806,6 +793,40 @@ func (r *HeatAPIReconciler) reconcileUpgrade(ctx context.Context, instance *heat
...
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.InputReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.InputReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, err
+	}
+
+	// Add a prefix to the var name to avoid accidental collision with other non-secret
```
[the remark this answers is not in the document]
zzzeek: looks like this merged before i got to change it :/

controllers/heat_controller.go
```diff
@@ -568,6 +568,22 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
...
+	// TODO(zzzeek) - It's not clear if this is called too early here.
+	// at the moment, heat_controller_test.go doesn't seem to have fixtures
+	// I can use to simulate getting all the way to the end of a reconcile
+	// for an instance.  Basically this should be called when any pods have
+	// been restarted to run on an updated set of DB credentials, and the old
+	// ones are no longer needed.  This would allow the scenario where
+	// a new MariaDBAccount is created and an old MariaDBAccount is marked
+	// deleted at once, where the finalizer will keep the old one around until
+	// it's safe to drop.
+	err = mariadbv1.DeleteUnusedMariaDBAccountFinalizers(ctx, helper, instance.Name, instance.Spec.DatabaseAccount, instance.Namespace)
```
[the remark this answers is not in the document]
zzzeek: so I have a suite of ginkgo tests that exercises the finalizers moving from one MariaDBAccount to another, once services are redeployed with an updated MariaDBAccount / username/password.    That's the suite thats added in the tests here.   It wont work if the DeleteUnusedMariaDBAccountFinalizers call above isn't called, since that's part of what it's testing.   

Right now, having the call where it is is *mostly* fine, you can rotate a new MariaDBAccount in and all will be fine.  The one thing you can't do here is delete the old MariaDBAccount before the new one is fully reconciled and running, without the risk of a brief race condition where existing pods are using the old database login that would be pulled out, since the finalizer here is removed slightly too early to be guaranteed of that.     The issue would of course be worked out within a few seconds in the average case and wouldn't be noticed, and we as yet dont have any system that is actually rotating out old/new MariaDBAccount objects, so there will be no impact from this for now.  However, the comment here should be kept that we would like to move this call downwards to be after we are ensured all pods have been re-deployed with new credentials.

PR #323, opened 2024-02-22
[tlse] TLS database connection

stuggi:
The my.cnf file gets added to the secret holding the service configs. The content of my.cnf is centrally managed in the mariadb-operator and retrieved calling db.GetDatabaseClientConfig(tlsCfg)

Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/190
Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/191

Jira: [OSPRH-4547](https://issues.redhat.com//browse/OSPRH-4547)

### DIFF
--- a/controllers/heatapi_controller.go
+++ b/controllers/heatapi_controller.go
@@ -55,6 +55,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
+	mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
 )
 
 // HeatAPIReconciler reconciles a Heat object
@@ -820,10 +821,22 @@ func (r *HeatAPIReconciler) generateServiceConfigMaps(
 
 	cmLabels := labels.GetLabels(instance, labels.GetGroupLabel(heat.ServiceName), map[string]string{})
 
+	db, err := mariadbv1.GetDatabaseByName(ctx, h, heat.DatabaseName)
+	if err != nil {
+		return err
+	}
+	var tlsCfg *tls.Service
+	if instance.Spec.TLS.Ca.CaBundleSecretName != "" {
+		tlsCfg = &tls.Service{}
+	}
+
 	// customData hold any customization for the service.
 	// custom.conf is going to /etc/heat/heat.conf.d
 	// TODO: make sure custom.conf can not be overwritten
-	customData := map[string]string{common.CustomServiceConfigFileName: instance.Spec.CustomServiceConfig}
+	customData := map[string]string{
+		common.CustomServiceConfigFileName: instance.Spec.CustomServiceConfig,
+		"my.cnf":                           db.GetDatabaseClientConfig(tlsCfg), //(mschuppert) for now just get the default my.cnf
+	}
 
 	for key, data := range instance.Spec.DefaultConfigOverwrite {
 		customData[key] = data
--- a/controllers/heat_controller.go
+++ b/controllers/heat_controller.go
@@ -25,6 +25,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/job"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
 
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/fields"
@@ -498,6 +499,13 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 
 	instance.Status.Conditions.MarkTrue(condition.RabbitMqTransportURLReadyCondition, condition.RabbitMqTransportURLReadyMessage)
 
+	db, result, err := r.ensureDB(ctx, helper, instance)
+	if err != nil {
+		return ctrl.Result{}, err
+	} else if (result != ctrl.Result{}) {
+		return result, nil
+	}
+
 	//
 	// Create ConfigMaps and Secrets required as input for the Service and calculate an overall hash of hashes
 	//
@@ -508,7 +516,7 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 	// - %-config configmap holding minimal heat config required to get the service up, user can add additional files to be added to the service
 	// - parameters which has passwords gets added from the OpenStack secret via the init container
 	//
-	err = r.generateServiceConfigMaps(ctx, instance, helper, &configMapVars, memcached)
+	err = r.generateServiceConfigMaps(ctx, instance, helper, &configMapVars, memcached, db)
 	if err != nil {
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			condition.ServiceConfigReadyCondition,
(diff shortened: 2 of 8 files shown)


controllers/heatapi_controller.go
```diff
@@ -820,10 +821,22 @@ func (r *HeatAPIReconciler) generateServiceConfigMaps(
 
 	cmLabels := labels.GetLabels(instance, labels.GetGroupLabel(heat.ServiceName), map[string]string{})
 
+	db, err := mariadbv1.GetDatabaseByName(ctx, h, heat.DatabaseName)
```
zzzeek, line 824: so in the patches I'm doing over at https://github.com/openstack-k8s-operators/mariadb-operator/pull/184 I've been explicitly moving *away* from using the Database object as a facade for *getting* things about the MariaDBDatabase and MariaDBAccount.  in my patches, in this function you will see this:

```
databaseAccount, dbSecret, err := mariadbv1.GetAccountAndSecret(ctx, h, instance.Spec.DatabaseAccount, instance.Namespace)
if err != nil {
	return err
}

// later
databaseConnection := fmt.Sprintf("mysql+pymysql://%s:%s@%s/%s",
			databaseAccount.Spec.UserName,
			string(dbSecret.Data[mariadbv1.DatabasePasswordSelector]),
			instance.Status.DatabaseHostname,
			heat.DatabaseName,
		)

```

this is because we need both the MariaDBAccount and the Secret to generate the mysql URL (which it also does here now, putting the URL in the config map since we dont want this in the job / pod template, and the configmap  **also** becomes a secret, not a configmap).     for heat you can see the patch at https://github.com/openstack-k8s-operators/heat-operator/pull/322/files 

I am not welded to moving away from Database, but I have done so because so far it does not feel well suited to the fact that we have at least *three* objects to deal with, MariaDBDatabase, MariaDBAccount, Secret, which makes it cumbersome to use as it either requires two separate arguments to have the correct data within it, or it is a "broken" state where it only has MariaDBDatabase and has to guess about MariaDBAccount.    if we are going to use the Database to get things then we need to set up the pattern for getting the Secret w/ the password as well.   if this merges the way it is, then I will have to fatten up this logic to also get the MariaDBAccount and probably the Secret all at the same time, otherwise we have a half-broken Database object that can't do all of the things it promises.

IOW GetDatabaseByName() does not work since it doesnt get you an account login anymore without guessing legacy things that are no longer correct once CRs have databaseAccount added.   the function is deprecated for this reason.

is there a reason we can't make this simpler and less intricate (and also more accurate, as GetDatabaseByName has to guess about the MariaDBAccount) by calling up the MariaDBDatabase directly?    

that is:

```
mariaDBDatabase, err := mariadbv1.GetMariaDBDatabase(ctx, h, heat.DatabaseName)
clientConfig := mariadbv1.GetDatabaseClientConfigs(mariaDBDatabase, tlsCfg)
```

otherwise can you suggest a better API for `.GetDatabaseByName` and the necessity to also have the MariaDBAccount / Secret?

would you want it like this?

```
db, err := mariadbv1.GetDatabaseByNameAndAccount(ctx, h, heat.DatabaseName, instance.Spec.DatabaseAccount)

clientConfig := db.GetDatabaseClientConfigs(tlsCfg)

dbSecret = db.GetSecret()  // or db.GetSecret(ctx, h)  , does an extra lookup
databaseAccount = db.GetAccount()

// later
databaseConnection := fmt.Sprintf("mysql+pymysql://%s:%s@%s/%s",
			databaseAccount.Spec.UserName,
			string(dbSecret.Data[mariadbv1.DatabasePasswordSelector]),
			instance.Status.DatabaseHostname,
			heat.DatabaseName,
		)

```

either way this has to change a bit

stuggi: > so in the patches I'm doing over at [openstack-k8s-operators/mariadb-operator#184](https://github.com/openstack-k8s-operators/mariadb-operator/pull/184) I've been explicitly moving _away_ from using the Database object as a facade for _getting_ things about the MariaDBDatabase and MariaDBAccount. in my patches, in this function you will see this:
> 
> ```
> databaseAccount, dbSecret, err := mariadbv1.GetAccountAndSecret(ctx, h, instance.Spec.DatabaseAccount, instance.Namespace)
> if err != nil {
> 	return err
> }
> 
> // later
> databaseConnection := fmt.Sprintf("mysql+pymysql://%s:%s@%s/%s",
> 			databaseAccount.Spec.UserName,
> 			string(dbSecret.Data[mariadbv1.DatabasePasswordSelector]),
> 			instance.Status.DatabaseHostname,
> 			heat.DatabaseName,
> 		)
> ```
> 
> this is because we need both the MariaDBAccount and the Secret to generate the mysql URL (which it also does here now, putting the URL in the config map since we dont want this in the job / pod template, and the configmap **also** becomes a secret, not a configmap). for heat you can see the patch at https://github.com/openstack-k8s-operators/heat-operator/pull/322/files
> 
> I am not welded to moving away from Database, but I have done so because so far it does not feel well suited to the fact that we have at least _three_ objects to deal with, MariaDBDatabase, MariaDBAccount, Secret, which makes it cumbersome to use as it either requires two separate arguments to have the correct data within it, or it is a "broken" state where it only has MariaDBDatabase and has to guess about MariaDBAccount. if we are going to use the Database to get things then we need to set up the pattern for getting the Secret w/ the password as well. if this merges the way it is, then I will have to fatten up this logic to also get the MariaDBAccount and probably the Secret all at the same time, otherwise we have a half-broken Database object that can't do all of the things it promises.

right, not have to deal with multiple objects was the reason for our discussion we had two weeks, or so, ago for using the mariadbdatabase to request a DB with accounts and get a ready secret with all the details to login to it (connection + my.cnf) and abstract the account rotation behind within that.

> 
> IOW GetDatabaseByName() does not work since it doesnt get you an account login anymore without guessing legacy things that are no longer correct once CRs have databaseAccount added. the function is deprecated for this reason.
> 
> is there a reason we can't make this simpler and less intricate (and also more accurate, as GetDatabaseByName has to guess about the MariaDBAccount) by calling up the MariaDBDatabase directly?
> 
> that is:
> 
> ```
> mariaDBDatabase, err := mariadbv1.GetMariaDBDatabase(ctx, h, heat.DatabaseName)
> clientConfig := mariadbv1.GetDatabaseClientConfigs(mariaDBDatabase, tlsCfg)
> ```

we can do that, it was was just not obvious to me that we going to move completely away from the Database. I kept the Account guessing like it is today until we land your work across the operators.

> 
> otherwise can you suggest a better API for `.GetDatabaseByName` and the necessity to also have the MariaDBAccount / Secret?
> 
> would you want it like this?
> 
> ```
> db, err := mariadbv1.GetDatabaseByNameAndAccount(ctx, h, heat.DatabaseName, instance.Spec.DatabaseAccount)
> 
> clientConfig := db.GetDatabaseClientConfigs(tlsCfg)
> 
> dbSecret = db.GetSecret()  // or db.GetSecret(ctx, h)  , does an extra lookup
> databaseAccount = db.GetAccount()
> 
> // later
> databaseConnection := fmt.Sprintf("mysql+pymysql://%s:%s@%s/%s",
> 			databaseAccount.Spec.UserName,
> 			string(dbSecret.Data[mariadbv1.DatabasePasswordSelector]),
> 			instance.Status.DatabaseHostname,
> 			heat.DatabaseName,
> 		)
> ```
> 
> either way this has to change a bit

either way is fine for me, we just have to agree and align across all the operators.

bshephar: Are we happy to move this conversation to a follow-up PR then? Seems like the intent of the PR is satisfied by the current PR and aligned with the pattern used in the other operators. So I'm happy to merge this and move any remaining design discussions to a follow up PR once we have agreed on this streamlined design.

bshephar (approved):
lgtm

PR #344, opened 2024-04-03
Fix HeatCnfAPI ReadyCondition check

fmount:
The previous patch missed the check of `AllSubConditionIsTrue` that, has a consequence, marks the global `ReadyCondition` as `True`. This patch fixes the wrong behavior adding the missing code block.


vyzigold:
/lgtm
I was about to start looking into it, but you were faster.

vyzigold:
Maybe it would be worth (in a different PR of course) if the kuttl tests checked the conditions?

abays (approved):
/lgtm

PR #345, opened 2024-04-04
Add observed generation for each controller

bshephar:
This change adds the required status.observedGeneration pieces for each of the heat-operator controllers. This is leveraged during the update workflow.


controllers/heatapi_controller.go
```diff
@@ -361,6 +361,8 @@ func (r *HeatAPIReconciler) reconcileDelete(ctx context.Context, instance *heatv
 	controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
 	r.Log.Info("Reconciled API delete successfully")
 
+	instance.Status.ObservedGeneration = instance.Generation
```
fmount, line 364: As per [1] we are trying to invert the pattern and bump `ObservedGeneration` at the beginning of the Reconciliation loop. This should be basically moved at the beginning of `reconcileNormal` [2] (more details in [1] and [3]).
The top level, as done in [1], should check the subResources ObservedGeneration before marking ReadyCondition as True.
Let' sync up and update this patch as long as [1] is finalized, so we can spread the pattern across the board.
(@dprince FYI ^) 

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/501
[2] https://github.com/openstack-k8s-operators/heat-operator/blob/main/controllers/heatapi_controller.go#L552
[3] https://github.com/openstack-k8s-operators/docs/pull/102

controllers/heatcfnapi_controller.go
```diff
@@ -365,6 +365,8 @@ func (r *HeatCfnAPIReconciler) reconcileDelete(ctx context.Context, instance *he
 	controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
 	r.Log.Info("Reconciled CfnAPI delete successfully")
 
+	instance.Status.ObservedGeneration = instance.Generation
```
fmount, line 368: same comment applies here

controllers/heatengine_controller.go
```diff
@@ -277,6 +277,8 @@ func (r *HeatEngineReconciler) reconcileDelete(ctx context.Context, instance *he
 	controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
 	r.Log.Info("Reconciled Engine delete successfully")
 
+	instance.Status.ObservedGeneration = instance.Generation
```
fmount, line 280: same comment applies here

controllers/heat_controller.go
```diff
@@ -691,6 +691,8 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 			condition.ReadyCondition, condition.ReadyMessage)
 	}
 	r.Log.Info("Reconciled Heat successfully")
+	instance.Status.ObservedGeneration = instance.Generation
```
fmount, line 694: This should go away and be bumped at the beginning of the `reconcileNormal`.  see next comment

fmount (changes requested):
Thank you for starting this, very helpful! As per [1] we're trying to consolidate a pattern that can be spread across the operators, so this patch might require a couple of changes, but I left some comments on the  `ObservedGeneration` bump that should happen at the beginning of the reconciliation loop. 

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/501/

controllers/heatapi_controller.go
bshephar: Sorry mate, apologies. I missed all of these comments, need a better email filtering system I think.

Ok, I have moved all of these now, did you want to take another look and make sure I have captured your intentions?

bshephar: Actually, now they're in the right place. So after the init of the other conditions.

controllers/heatengine_controller.go
bshephar: Moved to after the `Conditions.Init()`

controllers/heat_controller.go
bshephar: Moved to after the `Conditions.Init()`

fmount (approved):
/lgtm

PR #353, opened 2024-04-23
Check the ObservedGeneration for the generated subCRs

fmount:
This patch represents a follow up of the current `observedGeneration` work and the idea is to check that field in the sub resources and mirror the `Condition` (as well as `ReadyCount`) only when the generation is the last seen.
It represents a follow up of #345.

### DIFF
--- a/controllers/heat_controller.go
+++ b/controllers/heat_controller.go
@@ -623,16 +623,38 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 			err.Error()))
 		return ctrl.Result{}, err
 	}
-	if op != controllerutil.OperationResultNone {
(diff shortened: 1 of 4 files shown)


controllers/heat_controller.go
```diff
@@ -671,22 +715,40 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
...
+			instance.Status.Conditions.Set(c)
+		}
+		if op != controllerutil.OperationResultNone {
+			r.Log.Info(fmt.Sprintf("Deployment %s successfully reconciled - operation: %s", instance.Name, string(op)))
+		}
 	}
-
 	// We reached the end of the Reconcile, update the Ready condition based on
 	// the sub conditions
-	instance.Status.ObservedGeneration = instance.Generation
```
fmount, line 689: @bshephar a leftover from the previous patch :/ I removed it here

dprince (approved):

PR #361, opened 2024-05-08
Add webhook to validate service override endpoint type

stuggi:
Can also be called from the openstack-operator webhook on the HeatSpecCore, like for other operators in [1].

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/505

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/d2703d3a321c979dacaca95b5d4a634bf116e0db/apis/core/v1beta1/openstackcontrolplane_webhook.go#L181

### DIFF
--- a/api/v1beta1/heat_webhook.go
+++ b/api/v1beta1/heat_webhook.go
@@ -25,6 +25,7 @@ package v1beta1
 import (
 	"fmt"
 
+	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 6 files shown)


api/v1beta1/heat_webhook.go
```diff
@@ -99,10 +100,55 @@ var _ webhook.Validator = &Heat{}
...
-	// TODO(user): fill in your validation logic upon object creation.
+	var allErrs field.ErrorList
+	basePath := field.NewPath("spec")
+	if err := r.Spec.ValidateCreate(basePath); err != nil {
+		allErrs = append(allErrs, err...)
+	}
+
+	if len(allErrs) != 0 {
+		return nil, apierrors.NewInvalid(
+			schema.GroupKind{Group: "ironic.openstack.org", Kind: "Ironic"},
```
abays, line 111: Should this be `heat` and `Heat` instead of `ironic` and `Ironic`?

stuggi: right, good catch, thanks

abays (approved):
/lgtm

PR #362, opened 2024-05-08
Enable SRBAC by default

bshephar:
This change enables SRBAC by default for the Heat service.

### DIFF
--- a/api/bases/heat.openstack.org_heats.yaml
+++ b/api/bases/heat.openstack.org_heats.yaml
@@ -69,6 +69,11 @@ spec:
                   files. Those get added to the service config dir in /etc/<service>
                   . TODO: -> implement'
                 type: object
+              enableSecureRBAC:
+                default: true
+                description: EnableSecureRBAC - Enable Consistent and Secure RBAC
+                  policies
+                type: boolean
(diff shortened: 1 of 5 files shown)


mrunge:
Do we know heat actually works with SRBAC enabled? There is https://github.com/openstack-k8s-operators/keystone-operator/pull/395 and its revert https://github.com/openstack-k8s-operators/keystone-operator/pull/415

rabi:
heat would work with 

> Do we know heat actually works with SRBAC enabled? There is [openstack-k8s-operators/keystone-operator#395](https://github.com/openstack-k8s-operators/keystone-operator/pull/395) and its revert [openstack-k8s-operators/keystone-operator#415](https://github.com/openstack-k8s-operators/keystone-operator/pull/415)

That issues are not related to enabling SRBAC in heat. Heat kinda does not use scope in policies, so this PR is fine.  When keystone enforces scope, heat has issues as we internally need domain admin/domain user (created internally by heat) to list roles/access credentials using a domain scoped token.

bshephar:
We merged this one instead:
https://github.com/openstack-k8s-operators/heat-operator/pull/374

PR #372, opened 2024-06-05
Add default for DatabaseInstance

rabi:
Else it would be initialized to empty string when created with empty template and then webhook would not allow it to be modified.


bshephar (approved):
Sounds sane to me. Would be nice to get all of the service operators on-board with this one for consistency purposes.

PR #375, opened 2024-06-13
Handle spaces in encryption key

bshephar:
If a user tries to provide a auth_encryption_key containing spaces, the current implementation will only apply up until the first space. This will result in heat-engine failing to start because the encryption_key is not 32 characters or longer.


stuggi (approved):
/lgtm

PR #376, opened 2024-06-13
Set heat_metadata_server_url

bshephar:
This change adds the capability to detect and set the cfn route as the heat_metadata_server_url in the heat.conf file

### DIFF
--- a/controllers/heat_controller.go
+++ b/controllers/heat_controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
 
+	routev1 "github.com/openshift/api/route/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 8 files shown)


bshephar:
We need a `Watches()` here for the route. So we need to reconcile once the openstack-operator has created the CFN route, which will update the `heat.conf` file with the Route URL.

stuggi:
Can you explain why we need to do the route lookup? Don't we have already that info passed in via the `spec.heatCfnAPI.override.service.public.endpointURL`:

```
  spec:
  ...
    heatCfnAPI:
        override:
          service:
            ...
            public:
              endpointURL: https://heat-cfnapi-public-openstack.apps-crc.testing
```

which is also used to register the endpoint:
```
sh-5.1$ openstack endpoint list | grep heat
| 02e7a0925e1b426a963ef6f3ba1a7925 | regionOne | heat         | orchestration  | True    | public    | https://heat-api-public-openstack.apps-crc.testing/v1/%(tenant_id)s   |
| 7ea6768b31a144dcad25dbfe85017695 | regionOne | heat-cfn     | cloudformation | True    | public    | https://heat-cfnapi-public-openstack.apps-crc.testing/v1              |
| ba7dd48987274ac299afd1ec2cf53dad | regionOne | heat-cfn     | cloudformation | True    | internal  | https://heat-cfnapi-internal.openstack.svc:8000/v1                    |
| cb1d444f717644028767fd45e900f9b9 | regionOne | heat         | orchestration  | True    | internal  | https://heat-api-internal.openstack.svc:8004/v1/%(tenant_id)s         |
```

And if none is provided it should be just be the k8s service, same how we handle the endpoint registration.  Like the public apiEndpoint without the `dataPath` https://github.com/openstack-k8s-operators/heat-operator/blob/main/controllers/heatcfnapi_controller.go#L486C3-L486C37

bshephar:
> Can you explain why we need to do the route lookup? Don't we have already that info passed in via the `spec.heatCfnAPI.override.service.public.endpointURL`:
> 
> ```
>   spec:
>   ...
>     heatCfnAPI:
>         override:
>           service:
>             ...
>             public:
>               endpointURL: https://heat-cfnapi-public-openstack.apps-crc.testing
> ```
> 
> which is also used to register the endpoint:
> 
> ```
> sh-5.1$ openstack endpoint list | grep heat
> | 02e7a0925e1b426a963ef6f3ba1a7925 | regionOne | heat         | orchestration  | True    | public    | https://heat-api-public-openstack.apps-crc.testing/v1/%(tenant_id)s   |
> | 7ea6768b31a144dcad25dbfe85017695 | regionOne | heat-cfn     | cloudformation | True    | public    | https://heat-cfnapi-public-openstack.apps-crc.testing/v1              |
> | ba7dd48987274ac299afd1ec2cf53dad | regionOne | heat-cfn     | cloudformation | True    | internal  | https://heat-cfnapi-internal.openstack.svc:8000/v1                    |
> | cb1d444f717644028767fd45e900f9b9 | regionOne | heat         | orchestration  | True    | internal  | https://heat-api-internal.openstack.svc:8004/v1/%(tenant_id)s         |
> ```
> 
> And if none is provided it should be just be the k8s service, same how we handle the endpoint registration. Like the public apiEndpoint without the `dataPath` https://github.com/openstack-k8s-operators/heat-operator/blob/main/controllers/heatcfnapi_controller.go#L486C3-L486C37

Yeah, I too realised this after I put in all this effort...

I have another iteration of this in testing atm using the `endpointURL` value. But it seems this isn't going to fix the problem I was trying to fix in tempest jobs anyway. I had initially speculated that the Tempest jobs were failing because it couldn't reach the metadata endpoint (this heat_metadata_server_url). Which we set in tripleo here:
https://github.com/openstack-archive/tripleo-heat-templates/blob/stable/wallaby/deployment/heat/heat-engine-container-puppet.yaml#L174-L178

So I can use this CFN endpointURL to set that, but the actual problem I need to solve is the signal URL. It resolves it from Keystone as the internal URL and this parameter I'm trying to set here won't change that.

Anyway, I'll close this and come back to my other patch once I figure out the immediate problem.

PR #408, opened 2024-08-08
Add functional tests and HeatSpec webhook validation

bshephar:
We added a webhook validation for the HeatSpecCore struct in: https://github.com/openstack-k8s-operators/heat-operator/pull/407

However, we also need the same validation added to the HeatSpec ValidateUpdate function. This change adds the missing validation, along with a functional test to verify it works.


stuggi (approved):
/lgtm

PR #421, opened 2024-09-16
Refacter service secrets function

bshephar:
This change refactors the generateServiceSecrets function to deduplicate code and move logic related to vhosts and templateParameters into dedicated functions.

### DIFF
--- a/controllers/heat_controller_test.go
+++ b/controllers/heat_controller_test.go
@@ -1,0 +1,62 @@
+package controllers
+
+import (
+	"reflect"
+	"testing"
+
+	heatv1beta1 "github.com/openstack-k8s-operators/heat-operator/api/v1beta1"
+
+	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
+)
+
+func TestRenderVhost(t *testing.T) {
+	instanceTest1 := &heatv1beta1.Heat{}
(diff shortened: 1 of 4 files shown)


controllers/heat_controller_test.go
bshephar, line 1: I know this adds a new style of testing. But I feel like Heat operator is particularly difficult to test because of the Gophercloud dependency. So we either mock the entire thing, or setup an API server to send OpenStack API like responses. Both of those options feel like a lot of work, so I propose that we instead try to make our code more natively testable by splitting up logic into smaller functions and then using table driven tests to validate their functionality.

In this particular case, the table driven test allowed for me to pass TLS and Non TLS scenarios quickly, while in envtest we just validate the resulting secret data for one off those scenarios. I think if we take this same style and logic and apply it to the rest of the repo, we can improve our test coverage here without the additional burden of the OpenStack API responses.

fao89 (approved):

bshephar:
Kuttl is failing due to the issue with the latest go-toolset in Go 1.21. We'll need to wait for that to be fixed to recheck on this one.

stuggi:
we've put in a custom image in place, the issue should be gone

fao89 (approved):

PR #423, opened 2024-09-18
Set minimum Go version via env var

bshephar:
GOTOOLCHAIN to set 1.21.0 as the min required version and sets it in go.work file.


bshephar:
Should have been a commit on the Go1.21 PR, not it's own PR.

PR #424, opened 2024-09-19
Add customServiceConfigSecrets interface

fao89:
https://github.com/openstack-k8s-operators/dev-docs/blob/79f22c379fa681ac3de42dcb836d214ed8451438/service_config.md#render-multiple-secrets-customserviceconfigsecrets-interface

closes [OSPRH-10267](https://issues.redhat.com//browse/OSPRH-10267)


controllers/heat_controller.go
```diff
@@ -402,6 +407,20 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 	}
 	secretVars[ospSecret.Name] = env.SetValue(hash)
 
+	// check for custom config secrets
+	for _, secretName := range instance.Spec.CustomServiceConfigSecrets {
```
fmount, line 411: I'm wondering if we really need this check here: in theory we're not setting an hash based on `customServiceConfigSecret`, and if they do not exist, we return an err on L979 (from `generateConfigSecret`) and exit the reconciliation loop because "Config is not ready". That seems a more appropriate error for this missing input. 
However I don't think this is a blocker here, just a general comment on the fact we iterate over the same structure twice in the same reconciliation loop, getting (eventually) the same data as output.
@fao89 thoughts on this?

fao89: you are right, I'll drop it

templates/heat/bin/init.sh
```diff
@@ -45,7 +45,7 @@ done
 # and have it automatically detected, or would we have to somehow change the call
 # to the heat binary to tell it to use that custom conf dir?
 echo merging /var/lib/config-data/default/custom.conf into ${SVC_CFG_MERGED}
-crudini --merge ${SVC_CFG_MERGED} </var/lib/config-data/default/custom.conf
+crudini --merge ${SVC_CFG_MERGED} </var/lib/config-data/default/01-custom.conf
```
fmount, line 48: one interesting thing here (maybe for a follow up) would be to remove `crudini` and remove (entirely) the `initscript`.
- `00-default.conf` can be a static file, copied by kolla during `kolla_start` (see what I did here [1] [2]).
This way you can simplify a lot the startup logic and let kolla do both copy and merge for you in a `heat.conf.d` directory where your 00, 01, 02, 03 configs are stored.

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/templates/common/config/00-config.conf
[2] https://github.com/openstack-k8s-operators/glance-operator/blob/main/templates/glanceapi/config/glance-api-config.json

fmount (commented):
I have to check it better but this iteration looks ok.
There's code that can be simplified (e.g. removal of the init script) but I think it's out of scope here. For what concerns the `customServiceConfigSecrets` interface, lgtm

templates/heat/bin/init.sh
fmount: oh that's good, works for me then, I'm +1 to focus PRs on a smaller subset of changes.

fmount (approved):
/lgtm

bshephar (approved):

PR #451, opened 2024-10-23
Get keystone service project

fao89:
https://github.com/openstack-k8s-operators/keystone-operator/blob/b36564988843e6f9c1d29328a7651826ccf744de/controllers/keystoneservice_controller.go#L498-L507


controllers/heat_controller.go
```diff
@@ -1077,13 +1077,19 @@ func (r *HeatReconciler) ensureStackDomain(
...
+	}
+
 	// Create Heat user
 	userID, err := os.CreateUser(
 		r.Log,
 		openstack.User{
 			Name:      heat.StackDomainAdminUsername,
 			Password:  password,
-			ProjectID: "service",
+			ProjectID: project.ID,
```
rabi, line 1092: You don't need to provide project when creating domain users. You can just remove it.

PR #452, opened 2024-10-23
Don't create stack_domain_admin user with project

rabi:
This may result in a project scoped token issued that can't create projects in the domain i.e only a domain scoped token can create projects in the domain.


fao89:
I wonder if we should backport to 18.0-fr1

PR #453, opened 2024-10-24
Skip Heat tempest tests with known issues upstream

bshephar:
This change adds specific tempest tests that are known to not work upstream. Some issues for example include the lack of valid DNS server and the dependency on DNS to work with TLS endpoints that are enabled by default.

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -30,9 +30,36 @@
         heat_tempest_plugin.tests.scenario.test_server_cfn_init.CfnInitIntegrationTest.test_server_cfn_init
         heat_tempest_plugin.tests.scenario.test_aodh_alarm.AodhAlarmTest.test_alarm
         heat_tempest_plugin.tests.scenario.test_base_resources.BasicResourcesTest.test_base_resources_integration
+        heat_tempest_plugin.tests.scenario.test_server_software_config.SoftwareConfigIntegrationTest.test_server_software_config
+        heat_tempest_plugin.tests.scenario.test_octavia_lbaas.LoadBalancerTest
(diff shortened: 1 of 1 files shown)


zuul.d/jobs.yaml
```diff
@@ -30,9 +30,36 @@
...
         heat_tempest_plugin.tests.scenario.test_aodh_alarm.AodhAlarmTest.test_alarm
         heat_tempest_plugin.tests.scenario.test_base_resources.BasicResourcesTest.test_base_resources_integration
+        heat_tempest_plugin.tests.scenario.test_server_software_config.SoftwareConfigIntegrationTest.test_server_software_config
+        heat_tempest_plugin.tests.scenario.test_octavia_lbaas.LoadBalancerTest
+        heat_tempest_plugin.tests.scenario.test_aodh_alarm.AodhAlarmTest.test_alarm
         heat_tempest_plugin.tests.scenario.test_server_signal.ServerSignalIntegrationTest.test_server_signal_userdata_format_raw
         heat_tempest_plugin.tests.scenario.test_server_signal.ServerSignalIntegrationTest.test_server_signal_userdata_format_software_config
-        heat_tempest_plugin.tests.scenario.test_server_software_config.SoftwareConfigIntegrationTest.test_server_software_config
+        heat_integrationtests.functional.test_aws_stack
+        heat_integrationtests.functional.test_cancel_update.CancelUpdateTest.test_cancel_update_server_with_port
```
fao89, line 39: we don't have `heat_integrationtests` included here: https://github.com/openstack-k8s-operators/heat-operator/blob/main/zuul.d/jobs.yaml#L26-L27
should we add it?

PR #456, opened 2024-10-24
Refactor tempest job

fao89:


zuul.d/jobs.yaml
```diff
@@ -23,6 +22,8 @@
             validation.run_validation true
             identity.v3_endpoint_type public
             identity.v2_admin_endpoint_type public
+            heat_plugin.vm_to_heat_api_insecure true
+            heat_plugin.dns_servers "192.168.122.80"
```
bshephar, line 26: If you're thinking about using this to solve the signal test DNS issues, unfortunately this didn't work for me. Because this DNS server has the addresses for the k8s services, but not for the routes. Unless, you're also using metallb services with the k8s svc name as the signal endpoint I guess?

bshephar (approved):

fao89:
OSPRH-7385

fao89:
OSPRH-7384

PR #457, opened 2024-10-24
Fix server signal tests

fao89:


fao89:
recheck

zuul.d/jobs.yaml
```diff
@@ -23,15 +23,15 @@
             validation.run_validation true
             identity.v3_endpoint_type public
             identity.v2_admin_endpoint_type public
+            heat_plugin.vm_to_heat_api_insecure true
+            heat_plugin.dns_servers "192.168.122.80"
```
fao89, line 27: @bshephar even after adding this, the test still failing with:
```
curl: (6) Could not resolve host: heat-api-internal.openstack.svc
curl: (6) Could not resolve host: heat-api-internal.openstack.svc
curl: (6) Could not resolve host: heat-api-internal.openstack.svc
```
https://logserver.rdoproject.org/57/457/cc61e75c3cf6bc2de9a9bd90c00915fde264b84c/github-check/heat-operator-tempest-multinode/7e37094/controller/ci-framework-data/tests/test_operator/tempest-tests/stestr_results.html

conf:
https://logserver.rdoproject.org/57/457/cc61e75c3cf6bc2de9a9bd90c00915fde264b84c/github-check/heat-operator-tempest-multinode/7e37094/controller/ci-framework-data/tests/test_operator/tempest-tests/etc/tempest.conf

bshephar: It does seem like that DNS server should have answers for this query:
```
$ oc exec -it dnsmasq-dns-548bf5d8cb-lbr65 -- cat /etc/dnsmasq.d/hosts/dns-svc
Defaulted container "dnsmasq-dns" out of: dnsmasq-dns, init (init)
172.17.0.80 barbican-internal.openstack.svc cinder-internal.openstack.svc glance-default-internal.openstack.svc heat-api-internal.openstack.svc keystone-internal.openstack.svc neutron-internal.openstack.svc nova-internal.openstack.svc nova-metadata-internal.openstack.svc placement-internal.openstack.svc swift-internal.openstack.svc
172.17.0.85 rabbitmq.openstack.svc
172.17.0.86 rabbitmq-cell1.openstack.svc
```

Can a VM deployed on the EDPM node reach that DNS Servers IP address though? Like, that would require a OpenStack network created bridging the right interface on the EDPM compute node.

fao89: Good question! I don't know exactly how tempest job is configured but I believe we don't do that

PR #461, opened 2024-10-28
Index the customServiceConfigSecrets for use in the Secret lookup

bshephar:
This change ensures the customServiceConfigSecrets value is indexed for when we need to use it in the FindObjectsForSrc function


stuggi (approved):
/lgtm

PR #463, opened 2024-10-28
Add configurable API Timeouts

fao89:
This patch adds an apiTimeout field to the HeatSpecBase to allow human operators to simultaneously configure the timeouts for HAProxy, Apache, and the rpc_response_timeout.

The apiTimeout defaults to 60 seconds, to mimic the behavior present in OSP17.

Having different timeouts for HAProxy, Apache, and rpc_response_timeout is possible setting the HAProxy timeout in the apiOverride, the Apache timeout with apiTimeout, and setting rpc_response_timeout in the top level Heat customServiceConfig.

To be able to change the HAProxy value based on the apiTimeout with any update (and not just the first time) the code adds a custom annotation "api.heat.openstack.org/timeout" with the value that was initially set, this way flags it as being set by the heat-operator.

closes [OSPRH-10965](https://issues.redhat.com//browse/OSPRH-10965)

### DIFF
--- a/api/v1beta1/heat_webhook.go
+++ b/api/v1beta1/heat_webhook.go
@@ -238,3 +238,30 @@ func (r *Heat) ValidateDelete() (admission.Warnings, error) {
 	// TODO(user): fill in your validation logic upon object deletion.
 	return nil, nil
 }
+
+// SetDefaultRouteAnnotations sets HAProxy timeout values of the route
+// NOTE: it is used by ctlplane webhook on openstack-operator
+func (spec *HeatSpecCore) SetDefaultRouteAnnotations(annotations map[string]string) {
+	const haProxyAnno = "haproxy.router.openshift.io/timeout"
+	// Use a custom annotation to flag when the operator has set the default HAProxy timeout
+	// With the annotation func determines when to overwrite existing HAProxy timeout with the APITimeout
+	const heatAnno = "api.heat.openstack.org/timeout"
+
+	valHeat, okHeat := annotations[heatAnno]
+	valHAProxy, okHAProxy := annotations[haProxyAnno]
+
+	// Human operator set the HAProxy timeout manually
+	if !okHeat && okHAProxy {
+		return
+	}
+
+	// Human operator modified the HAProxy timeout manually without removing the Heat flag
+	if okHeat && okHAProxy && valHeat != valHAProxy {
+		delete(annotations, heatAnno)
+		return
+	}
+
+	timeout := fmt.Sprintf("%ds", spec.APITimeout)
+	annotations[heatAnno] = timeout
+	annotations[haProxyAnno] = timeout
+}
--- a/controllers/heat_controller.go
+++ b/controllers/heat_controller.go
@@ -1353,6 +1353,7 @@ func initTemplateParameters(
 		"MemcachedServersWithInet": mc.GetMemcachedServerListWithInetString(),
 		"MemcachedTLS":             mc.GetMemcachedTLSSupport(),
 		"DatabaseConnection":       mysqlConnectionString,
+		"Timeout":                  instance.Spec.APITimeout,
 	}
 }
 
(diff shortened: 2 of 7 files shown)


api/v1beta1/heat_webhook.go
```diff
@@ -238,3 +238,29 @@ func (r *Heat) ValidateDelete() (admission.Warnings, error) {
...
+
+	// Human operator modified the HAProxy timeout manually without removing the Heat flag
+	if okHeat && okHAProxy && valHeat != valHAProxy {
+		delete(annotations, heatAnno)
+		return
+	}
+
+	timeout := fmt.Sprintf("%ds", spec.APITimeout)
+	annotations[heatAnno] = timeout
+	annotations[haProxyAnno] = timeout
```
bshephar, line 265: How are these being used? This function isn't called from anywhere that I can see, and it doesn't return anything. Should we be taking in a pointer to an annotations map instead?

fao89: I'm following the pattern of calling it from openstack-operator webhook:
- https://github.com/openstack-k8s-operators/manila-operator/blob/main/api/v1beta1/manila_webhook.go#L241
- https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/core/v1beta1/openstackcontrolplane_webhook.go#L783
- https://github.com/search?q=org%3Aopenstack-k8s-operators%20SetDefaultRouteAnnotations&type=code

after this PR gets merged, I'll open a PR against openstack-operator

fao89: https://github.com/openstack-k8s-operators/openstack-operator/pull/1163

templates/heat/config/00-default.conf
```diff
@@ -7,6 +7,8 @@ num_engine_workers=4
 auth_encryption_key={{ .AuthEncryptionKey }}
 use_stderr=true
 transport_url={{ .TransportURL }}
+# Keep the RPC call timeout in sync with HAProxy and Apache timeouts
```
rabi, line 10: I don't think `rpc_response_timeout` default of `60` would work for heat. It's `600` in 17.1 along with apache/httpd timeout. Btw where is haproxy in the picture?

fao89: it is the route timeout: https://docs.openshift.com/container-platform/4.17/networking/routes/route-configuration.html#nw-configuring-route-timeouts_route-configuration

rabi: right, should we mention the route timeout instead of haproxy though it translates to that?

fao89: I'll change it

bshephar: HAProxy as in the one used for OpenShift ingress. So we pass an annotation to the Route to configure HAProxy timeout on the OpenShift Route.

Does it make sense to group RPC with API? Feels like a loose relationship at best.

fao89: from [OSPRH-10954](https://issues.redhat.com/browse/OSPRH-10954), I tried to follow what manila, cinder, glance did. And I saw that both manila and cinder grouped it:
- https://github.com/openstack-k8s-operators/manila-operator/blob/57ac502c086d0bb2092e124d3152ab9e31a4ad79/templates/manila/config/00-config.conf#L13
- https://github.com/openstack-k8s-operators/cinder-operator/blob/980b1ab26c714bbc0b7f4925517c68f8d3e53169/templates/cinder/config/00-global-defaults.conf#L32

controllers/heat_controller.go
```diff
@@ -1353,6 +1353,7 @@ func initTemplateParameters(
 		"MemcachedServersWithInet": mc.GetMemcachedServerListWithInetString(),
 		"MemcachedTLS":             mc.GetMemcachedTLSSupport(),
 		"DatabaseConnection":       mysqlConnectionString,
+		"TimeOut":                  instance.Spec.APITimeout,
```
olliewalsh, line 1356: nit: Timeout used everywhere else

rabi (commented):
/lgtm

PR #494, opened 2025-02-10
Consume Topology CR by reference

fmount:
This patch provides more granular control over `Pod` placement and scheduling through the `Topology CR` integration introduced in infra-operator.
In particular it provides:

- a new API parameter (`TopologyRef`) defined for each `Component` that allows to reference an existing `Topology CR` in the same namespace

- the operator logic that retrieves and processes the referenced `Topology CR` through the functions provided by `lib-common`

- enhanced `Deployment` configuration that incorporates the processed `Topology`

- `kuttl` test to verify a referenced `Topology` has been applied to the `Deployment` and the resulting `Pods`

Note that `webhooks` are in place to prevent referencing a `Topology` from a different namespace (which is not a supported scenario), and this part is covered by envTests.


fmount:
Like we did for `mariadb`, we probably need to patch `install_yamls` to deploy `infra-operator` as a common step, because it provides a shared resource (Topology CRD) that needs to be installed in a kuttl environment.
For this reason I expect the kuttl tests to fail, and I'll follow up w/ the install_yamls patch soon.

fmount:
> Like we did for `mariadb`, we probably need to patch `install_yamls` to deploy `infra-operator` as a common step, because it provides a shared resource (Topology CRD) that needs to be installed in a kuttl environment. For this reason I expect the kuttl tests to fail, and I'll follow up w/ the install_yamls patch soon.

This one uses openstack-operator and all the CRDs are available when kuttl is run, so we're good:
```
--- PASS: kuttl (172.10s)
    --- PASS: kuttl/harness (0.00s)
        --- PASS: kuttl/harness/heat_scale (92.41s)
        --- PASS: kuttl/harness/heat_topology (41.52s)
        --- PASS: kuttl/harness/heat_tls (37.60s)
```

stuggi (approved):
/lgtm

PR #498, opened 2025-02-17
Tls tidy

bshephar:
Move mounts and volumes to dedicated function. This declutters the Deployment logic and makes each component easier to test. Particularly for Heat where we have some challenges with EnvTest and mocking the OpenStack API.

- Move Volume and VolumeMount creation to dedicated function
- Move TLS handling to dedicated functions for HeatAPI and HeatCfnAPI
- Return errors from HeatEngine deployment creation process if any formatting fails.
- Handle errors in the reconcile flow for HeatEngine controller.

### DIFF
--- a/pkg/heatapi/deployment.go
+++ b/pkg/heatapi/deployment.go
@@ -61,34 +61,13 @@ func Deployment(
 	volumes = append(volumes, secretVolumes...)
 	volumeMounts = append(volumeMounts, secretMounts...)
 
-	// add CA cert if defined
-	if instance.Spec.TLS.CaBundleSecretName != "" {
-		volumes = append(volumes, instance.Spec.TLS.CreateVolume())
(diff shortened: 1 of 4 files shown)


fao89:
recheck

pkg/heatapi/deployment.go
```diff
@@ -159,3 +137,58 @@ func Deployment(
 	}
 	return deployment, nil
 }
+
+func createVolumeAndMounts(instance *heatv1beta1.HeatAPI) ([]corev1.Volume, []corev1.VolumeMount, error) {
```
bshephar, line 141: Ok, I think this is working now. So, ideally, I would like these to be generic enough that we can use them for each deployment without reimplementing and duplicating the code.

Moving back to draft while I work on that.

bshephar: https://github.com/openstack-k8s-operators/heat-operator/pull/503

bshephar:
recheck

PR #501, opened 2025-02-24
Use common security context

bshephar:
This change uses the common SecurityContext and in turn sets RunAsRoot to False for all Heat related deployments.


bshephar:
```
$ oc logs heat-cfnapi-d7cf996cc-65hlm
+ sudo -E kolla_set_configs
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
```

stuggi (approved):
/lgtm

PR #503, opened 2025-02-26
Use common GetVolumes() GetVolumeMounts() funcs

bshephar:
Each service is duplicating logic to get Volumes and VolumeMounts. This change moves all services over to the top level implementations and makes the necessary compatibility changes to the top level functions.


abays:
Looks like KUTTL might need some tweaking to align with the changes:

```
case.go:380: resource Deployment:heat-kuttl-tests/heat-engine: .spec.template.spec.containers.volumeMounts.mountPath: value mismatch, expected: /var/lib/config-data/default != actual: /var/lib/kolla/config_files/config.json
```

pkg/heat/dbsync.go
```diff
@@ -40,7 +40,7 @@ func DBSyncJob(
 	envVars["KOLLA_CONFIG_STRATEGY"] = env.SetValue("COPY_ALWAYS")
 	envVars["KOLLA_BOOTSTRAP"] = env.SetValue("true")
 
-	volumes := GetVolumes(ServiceName)
+	volumes := GetVolumes(ServiceName, ServiceName)
```
fmount, line 43: Should this be: `GetVolumes(ServiceName, instance.Name)` ?

I understand that in case of `dbsync` you only need parent (=heat) volumes, but in case (that probably won't never happen) the CR has a different name, you might fail passing the secret.
This [1] is an example to give you the idea of what I was just describing.

[1] https://github.com/openstack-k8s-operators/manila-operator/blob/main/pkg/manilaapi/statefulset.go#L71

bshephar: Yeah, seems reasonable. Updated now.

PR #521, opened 2025-04-22
Make sure we reconcile heat-engine secrets

bshephar:
We trigger reconciles for each controllers based on different events. This can lead to a situation where Heat engine - which doesn't have the same number of events that will trigger a reconcile, becomes out of sync with the other controllers. For example, the configs which are essentially the same but updated with https endpoints can be synced by the HeatAPI and HeatCfnApi controllers first, before the HeatEngine controller finally catches up. In this interval of time the heat-engine process will fail since it's trying to speak to non https endpoints instead of the correct https ones.

Jira: https://issues.redhat.com/browse/OSPCIX-795

### DIFF
--- a/controllers/heatapi_controller.go
+++ b/controllers/heatapi_controller.go
@@ -257,7 +257,7 @@ func (r *HeatAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


stuggi (approved):
/lgtm

PR #532, opened 2025-05-15
Add extraMounts interface

rabi:
We can inject policy files, heat plugins etc using this.

Jira: https://issues.redhat.com/browse/OSPRH-16695


olliewalsh:
Also should inherit the top-level mounts, similar to https://github.com/openstack-k8s-operators/openstack-operator/pull/1441

rabi:
> Also should inherit the top-level mounts, similar to [openstack-k8s-operators/openstack-operator#1441](https://github.com/openstack-k8s-operators/openstack-operator/pull/1441)

You mean making additional change in openstack-operator right?

rabi:
recheck

rabi:
recheck

fmount:
The patch itself look good to me. I'm wondering if we might want to add an `envTest` to validate the new spec parameter like we did in [1] (the same test is present in the other operators implementing `extraMounts`).

We can do in a follow up, but let me know if you want to have an additional commit here before we land this patch.

[1] https://github.com/openstack-k8s-operators/manila-operator/blob/main/test/functional/manila_controller_test.go#L1115

rabi:
> We can do in a follow up, but let me know if you want to have an additional commit here before we land this patch

Looks like heat-operator envTest coverage is limited and I don't see anything from deployments checked in any of the tests. It would require some work to assert that volume and mounts are propagated for heat-engine and api deployments. I'll add tests in a followup as this has passed CI and ready to be merged.

fmount (approved):
/lgtm

PR #643, opened 2026-04-23
Add application credential finalizer management

Deydra71:
Jira: [OSPRH-29269](https://redhat.atlassian.net/browse/OSPRH-29269)

Application Credential dev-doc: https://github.com/openstack-k8s-operators/dev-docs/blob/main/application_credentials.md

* Tracks the active AC secret name in `Status.ApplicationCredentialSecret`
* Add `openstack.org/heat-ac-consumer` finalizer to the AC secret after service config is rendered
* On AC rotation, move the finalizer from the old secret to the new one
* On CR deletion, remove the consumer finalizer from the AC secret before cleaning up the CR

This ensures that the keystone-operator cannot revoke a rotated AC secret while Heat is still consuming it.

```
2026-04-28T11:55:51Z	INFO	Controllers.Heat	Added consumer finalizer	{"controller": "heat", "controllerGroup": "heat.openstack.org", "controllerKind": "Heat", "Heat": {"name":"heat","namespace":"openstack"}, "namespace": "openstack", "name": "heat", "reconcileID": "82672178-a3a3-4bd4-b3d5-87c6d5853d9b", "object": "ac-heat-aa0a4-secret", "finalizer": "openstack.org/heat-ac-consumer"}
2026-04-28T11:55:51Z	INFO	Controllers.Heat	Removed consumer finalizer	{"controller": "heat", "controllerGroup": "heat.openstack.org", "controllerKind": "Heat", "Heat": {"name":"heat","namespace":"openstack"}, "namespace": "openstack", "name": "heat", "reconcileID": "82672178-a3a3-4bd4-b3d5-87c6d5853d9b", "object": "ac-heat-ba179-secret", "finalizer": "openstack.org/heat-ac-consumer"}
```

Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/685

Assisted-by: Claude Opus 4.6 [noreply@anthropic.com](mailto:noreply@anthropic.com)

### DIFF
--- a/api/v1beta1/heat_types.go
+++ b/api/v1beta1/heat_types.go
@@ -42,7 +42,6 @@ const (
 	HeatDatabaseMigrationAnnotation = "heat.openstack.org/database-migration"
 )
 
-
 // HeatSpec defines the desired state of Heat
 type HeatSpec struct {
 	HeatSpecBase `json:",inline"`
@@ -176,6 +175,12 @@ type HeatStatus struct {
 
 	// ReadyCount of Heat Engine instance
 	HeatEngineReadyCount int32 `json:"heatEngineReadyCount,omitempty"`
+
+	// ApplicationCredentialSecret - the AC secret Heat is currently
+	// consuming and protecting with the openstack.org/heat-ac-consumer
+	// finalizer. Tracked so the controller can remove its finalizer from the
+	// old secret when the openstack-operator rotates the reference.
+	ApplicationCredentialSecret string `json:"applicationCredentialSecret,omitempty"`
 
 	// ObservedGeneration - the most recent generation observed for this
 	// service. If the observed generation is less than the spec generation,
--- a/internal/controller/heat_controller.go
+++ b/internal/controller/heat_controller.go
@@ -466,6 +466,19 @@ func (r *HeatReconciler) reconcileDelete(ctx context.Context, instance *heatv1be
 		}
 	}
 
+	// Remove consumer finalizer from AC secrets Heat was consuming.
+	// Check both status and spec to handle the edge case where the reconciler
+	// crashed after adding the finalizer but before updating the status.
+	for _, secretName := range []string{
+		instance.Status.ApplicationCredentialSecret,
+		instance.Spec.Auth.ApplicationCredentialSecret,
+	} {
+		if err := keystonev1.RemoveACSecretConsumerFinalizer(ctx, helper, instance.Namespace,
+			secretName, heat.ACConsumerFinalizer); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
 	// Service is deleted so remove the finalizer.
 	controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
 	Log.Info("Reconciled Heat delete successfully")
@@ -813,6 +826,24 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 
 	// Create Secrets - end
 
+	// Add consumer finalizer to the new AC secret early, before deployment.
+	// The old secret's finalizer is removed later (after all services deploy)
+	// so that rapid rotations don't revoke a credential still in use by pods.
+	if instance.Spec.Auth.ApplicationCredentialSecret != "" {
+		if err := keystonev1.ManageACSecretFinalizer(ctx, helper, instance.Namespace,
+			instance.Spec.Auth.ApplicationCredentialSecret,
+			"",
+			heat.ACConsumerFinalizer); err != nil {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				condition.ServiceConfigReadyCondition,
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				condition.ServiceConfigReadyErrorMessage,
+				err.Error()))
+			return ctrl.Result{}, err
+		}
+	}
+
 	instance.Status.Conditions.MarkTrue(condition.ServiceConfigReadyCondition, condition.ServiceConfigReadyMessage)
 
 	//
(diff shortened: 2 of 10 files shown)


internal/controller/heat_controller.go
```diff
@@ -466,6 +466,19 @@ func (r *HeatReconciler) reconcileDelete(ctx context.Context, instance *heatv1be
 		}
 	}
 
+	// Remove consumer finalizer from AC secrets Heat was consuming.
+	// Check both status and spec to handle the edge case where the reconciler
+	// crashed after adding the finalizer but before updating the status.
+	for _, secretName := range []string{
+		instance.Status.ApplicationCredentialSecret,
+		instance.Spec.Auth.ApplicationCredentialSecret,
```
bshephar, line 474: Where is `Auth` coming from? The HeatSpec doesn't have a `Auth` defined. Should this change also be adding `Auth`?

internal/controller/heat_controller.go
```diff
@@ -813,6 +826,23 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 
 	// Create Secrets - end
 
+	// Manage consumer finalizer, the AC data was already read and rendered to the service config secret
+	if instance.Spec.Auth.ApplicationCredentialSecret != "" || instance.Status.ApplicationCredentialSecret != "" {
+		if err := keystonev1.ManageACSecretFinalizer(ctx, helper, instance.Namespace,
+			instance.Spec.Auth.ApplicationCredentialSecret,
+			instance.Status.ApplicationCredentialSecret,
```
bshephar: Just a note that `Status.ApplicationCredentialSecret` is set to `omitempty`, so if this is undefined for any reason, you will be passing `nil` here which could end up with nil pointer deref issues. I would either check that `instance.Status.ApplicationCredentialSecret != nil` before doing this, or if you think it will always be set, remove the `omitempty` from the Status field?

bshephar (changes requested):
I'm an outsider now, as an outsider it would be nice to have some more context about this change in the commit message. :)

Maybe it's still a WIP, but there's a few important things I've noted in-line.

internal/controller/heat_controller.go
Deydra71: Hi @bshephar! Thanks for reviewing.

`Auth` comes from HeatTemplate (embedded inline in HeatSpecBase -> HeatSpec), defined in common_types.go  https://github.com/openstack-k8s-operators/heat-operator/blob/main/api/v1beta1/common_types.go#L63 .  So, it already comes from earlier change.

internal/controller/heat_controller.go
Deydra71: `ApplicationCredentialSecret` is a string type, so its zero value is "", not nil, as far as I see there's no nil pointer dereference risk here. And the != "" check on the line above already guards against passing an empty string to the finalizer function

Deydra71:
recheck

Deydra71:
Following the [discussion in watcher-operator](https://github.com/openstack-k8s-operators/watcher-operator/pull/369#discussion_r3234900678) the AC finalizer management is now split into two phases:

* Early phase: adds consumer finalizer to the new AC secret immediately (protects it from premature revocation)
* Late phase: removes consumer finalizer from the old AC secret only after `AllSubConditionIsTrue()` (all sub-services deployed with new credentials)

This prevents a race condition where rapid AC rotations could revoke credentials still in use by running pods.

The new file `api_fixture.go`:  heat's functional tests need a fake Keystone HTTP server to satisfy Keystone calls. SImilar pattern is used e.g. in octavia-operator - https://github.com/openstack-k8s-operators/octavia-operator/blob/main/test/functional/api_fixture.go

internal/controller/heat_controller.go
bshephar: I think my fork was out of sync. Yeah, I see it there, my bad.

api/v1beta1/heat_types.go
```diff
@@ -177,6 +176,12 @@ type HeatStatus struct {
 	// ReadyCount of Heat Engine instance
 	HeatEngineReadyCount int32 `json:"heatEngineReadyCount,omitempty"`
 
+	// ApplicationCredentialSecret - the AC secret Heat is currently
+	// consuming and protecting with the openstack.org/heat-ac-consumer
+	// finalizer. Tracked so the controller can remove its finalizer from the
+	// old secret when the openstack-operator rotates the reference.
+	ApplicationCredentialSecret string `json:"applicationCredentialSecret,omitempty"`
```
rabi, line 183: Are we backing up and restoring status for CRs? f not, the controller loses the previous applicationCredentialSecret reference, so stale consumer finalizers on older rotated app-cred Secrets may never get cleaned up.

internal/controller/heat_controller.go
```diff
@@ -813,6 +826,24 @@ func (r *HeatReconciler) reconcileNormal(ctx context.Context, instance *heatv1be
 
 	// Create Secrets - end
 
+	// Add consumer finalizer to the new AC secret early, before deployment.
```
rabi, line 829: Some of these comments explain the code in too much detail. We should avoid verbose comments that restate what the code is doing and keep comments focused on the key invariant or non-obvious why.

api/v1beta1/heat_types.go
Deydra71: No, only the usual reconcile time save/restore of conditions, not this field. But if a stale finalizer situation happens it only blocks Kubernetes cleanup - it does not extend the appcred in Keystone. The AC still expires on its configured `expirationDays` / `expiresAt`

internal/controller/heat_controller.go
Deydra71: THe comments are unified across all service operators. Is this a blocking change? Meaning should I push update reducing the verbosity?

api/v1beta1/heat_types.go
rabi: No, I am talking about CR backup and restore. I think we may end up with stale app cred resources as heat won't cleanup the consumer finalizers for them. But that's  probably a  limiation across all CRs with this approach and we have  live with it.

internal/controller/heat_controller.go
rabi: Comments unifed across operators? :smile:  We should probably clean it up as it makes code unreadable/unmaintainable in the long run (AI also generates lots of comments unnecessarily). But, I won't block on it.

Deydra71: I agree that going forward with agentic code can be problematic, the original comments were even more robust... Other reviewers didn't raise this issue so far, finding a common ground is key. But I don't think that having similar comments in different operators is an issue, since the maintainers are usually different. But something to keep in mind in general, yes.

api/v1beta1/heat_types.go
Deydra71: Ah, I see now - you're right, this is a limitation across service operators utilizing app creds

Deydra71:
@rabi Hi! Is there still any requested change/clarification on your side, or we ready to merge?

rabi (approved):

Deydra71:
/test heat-operator-build-deploy-kuttl

failed during install_yamls deploy

